Issue #19, opened 2020-08-14
Running yarn test throws an authentication error 


guilhermedeandrade:
I've successfully ran `yarn db:setup` and I have the `MyBysonApp_dev` database created with some tables such as Profile and User. However, when I try to run `yarn test` this error is thrown:

```sh
  ● Test suite failed to run

    Command failed: yarn prisma migrate up --create-db --experimental
    Environment variables loaded from prisma/.env
    Error: P1000: Authentication failed against database server at `localhost`, the provided database credentials for `postgres` are not valid.

    Please make sure to provide valid database credentials for the database server at `localhost`.
    error Command failed with exit code 1.
```

I've already changed both `prisma/.env` and `.env.test` files and added the user and password combination for my local postgres server.

Also, I did not face any issue by running `yarn prisma migrate up --create-db --experimental`

cball:
@guilhermedeandrade if migrate works,  the `DATABASE_URL` value in `prisma/.env` is correct. Can you confirm the `DATABASE_URL` in `.env.test` is the same value?

guilhermedeandrade:
The only difference is the name of the DB, which is either MyBisonApp_dev or MyBisonApp_test @cball 

`prisma/.env`: DATABASE_URL="postgresql://postgres:docker@localhost:5432/MyBisonApp_dev?schema=public"

`.env.test`
DATABASE_URL="postgresql://postgres:docker@localhost:5432/MyBisonApp_test?schema=public"

cball:
thanks @guilhermedeandrade. I imagine we'll need to improve the setup a bit to handle different database configurations. We'll try and reproduce this and get back to you.

Is that postgres user a superuser?

guilhermedeandrade:
> thanks @guilhermedeandrade. I imagine we'll need to improve the setup a bit to handle different database configurations. We'll try and reproduce this and get back to you.
> 
> Is that postgres user a superuser?

Yes, it is:

![image](https://user-images.githubusercontent.com/8797405/90341761-2099f480-dffa-11ea-8f5c-f1c8f9c6a6e6.png)

cball:
@guilhermedeandrade this may be the culprit: https://github.com/echobind/bisonapp/blob/master/template/tests/jest.setup.js#L18.

Could you try changing that URL to `postgresql://postgres:docker@localhost:5432/MyBisonApp_test?schema=public` locally and see if that allows tests to run? If so, we'll work on a fix for it.

[...]

PR #29, opened 2020-08-19
fix: schema value referenced is null resulting in the schema not being dropped during jest teardown

isaiahgrey93:
Schema value reference in jest setup script is null/undefined. This results in the teardown script failing to drop the temporary test schema.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,18 @@
+Description of PR that completes issue here...
+
+## Changes
+
+- Description of changes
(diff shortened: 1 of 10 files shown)


cball:
@isaiahgrey93 great catch. It seems this is different, yet related to #19. Should we add both fixes to this PR?

cball:
Let's bring this into `next` and add a followup for the other issue in #19.

isaiahgrey93:
> Let's bring this into `next` and add a followup for the other issue in #19.

Added with https://github.com/echobind/bisonapp/pull/34

PR #34, opened 2020-08-20
feat: generate the db url based on the .env DATABASE_URL

isaiahgrey93:
## Changes

- Generate the db url based on the .env DATABASE_URL by manipulating the path + schema param with node.js `URL`

## Screenshots

(prefer animated gif)

## Checklist


template/tests/jest.setup.js
```diff
@@ -13,13 +13,20 @@ module.exports = async () => {
...
-  global.schema = `test_${nanoid().toLowerCase()}`;
-  global.databaseUrl = `postgresql://postgres:postgres@localhost:5432/testing?schema=${this.schema}`;
+  global.schema = testSchema;
+  global.databaseUrl = testDatabaseUrl;
 
   process.env.DATABASE_URL = global.databaseUrl;
   global.process.env.DATABASE_URL = global.databaseUrl;
 
+  console.log(process.env.DATABASE_URL);
+
```
isaiahgrey93, line 29:
```suggestion
```

template/tests/jest.setup.js
```diff
@@ -13,9 +13,14 @@ module.exports = async () => {
   //   isUp = await statusCheck();
   // }
   // console.log(`[jest] server is up...`);
+  const testDatabaseName = 'testing';
+  const testSchema = `test_${nanoid().toLowerCase()}`;
+  const testDatabaseUrl = new URL(process.env.DATABASE_URL);
+  testDatabaseUrl.pathname = `/${testDatabaseName}`;
```
isaiahgrey93, line 19: Do we need to modify the database used @cball? Or will the different schema suffice?

cball: The only potential downside I can think of here is if someone creates a postgres user that is not a super user or does not have `create schema` permissions.

We either need to require that the postgres user has the right permissions, or we should take `DATABASE_URL` wholesale if defined.

I'm leaning towards requiring a postgres user with the right permissions and keeping the auto generated schema. what do you think?

PR #38, opened 2020-08-25
feat(deployment): support Heroku

### LINKED ISSUE #30: Support Heroku / Non-Serverless Deployments
The bulk of the changes should be confined to `pages/api/graphql/index`, but we'll need to make some adjustments for deployment and ask the user if they want a serverless deployment or not.

Heroku is definitely not a typical Jamstack host, but because we recommend it for databases, and it sometimes need to be used, we want Bison to support it.

cball:
This allows deploying to Heroku. It includes everything from server and database config to pipelines.

## Changes

- dont format ejs files with prettier
- update README
- adds inquirer to prompt user for values
- kicks off the workflow by asking user to past the git url
- deploy to Heroku
  - verify login
  - optionally create apps, databases, and pipelines
  - update postinstall message to use chalk and give better direction
- split docs into ci, deployment, workflow, faq, etc
- reworks simple template strings to use ejs
- pass chalk function to ejs templates to render colors and bold
- add ejs-lint dependency to lint templates (will be used in future tests as well)

## Screenshots
New CLI:
image

Resulting pipeline:
image

Gif of questions:

## Checklist

Fixes #30


template/.github/workflows/main.js.yml.ejs
```diff
@@ -140,3 +158,10 @@ jobs:
           # SET THIS OR YOU WILL DEPLOY TO YOUR PERSONAL VERCEL NAMESPACE
```
isaiahgrey93, line 158: Is it clear what property we need to set? Can we add an indicator to make sure they understand which line above or below we are indicating with this comment.
```suggestion
          # SET THIS👇 OR YOU WILL DEPLOY TO YOUR PERSONAL VERCEL NAMESPACE
```

cball: good point. we should also set a followup to bring this into the CLI prompts when choosing Vercel

PR #64, opened 2020-09-30
chore: extend the dev workflow doc with examples

alvincrespo:
[rendered](https://github.com/echobind/bisonapp/blob/chore/update-dev-workflow-doc/docs/devWorkflow.md)

This PR extends the dev workflow doc to include 1) examples for the API and 2) add the Prisma schema step.

## Changes

- Adds example utilizing the real-world example of `Organization` type
- Includes Prisma schema step

## Checklist


cball (approved):
👍  this reads so much better.

PR #65, opened 2020-09-30
fix: replace queryType with extendType

alvincrespo:
Nexus will error out if a previous `queryType` is already defined. We're planning to include a "global" that sets this - so ensuring that generated modules just extend Query will ensure this error does not happen for a dev.

## Changes

- Replaces `queryType` in our generated template with `extendType`

## Checklist

### DIFF
--- a/template/_templates/graphql/new/graphql.ejs
+++ b/template/_templates/graphql/new/graphql.ejs
@@ -27,7 +27,8 @@ export const CallPreference = enumType({
 });
 
 // Queries
-schema.queryType({
+schema.extendType({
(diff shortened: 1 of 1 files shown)


template/_templates/graphql/new/graphql.ejs
```diff
@@ -27,7 +27,8 @@ export const CallPreference = enumType({
 });
 
 // Queries
-schema.queryType({
+schema.extendType({
```
cball, line 30: 🎸

cball (approved):

PR #67, opened 2020-10-01
feat: add support for dropping the database

alvincrespo:
This PR ensures that a developer can drop their database from the script `db:drop`

## Changes

- Adds the script `db:drop` to a generated apps `package.json`
- Includes Mocha for unit testing Bison
- Creates an `createApp` script to create the app but NOT start it (to verify setup)
- Defines a unit test to ensure that `DATABASE_NAME` is properly generated in `prisma/.env`
- Extends our `Release` workflow to include unit tests
- Creates a `test:prisma` script for independently testing prisma setup

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,14 @@ jobs:
       - name: Install packages
         run: |
           yarn install --frozen-lockfile
+      - name: Run Unit Tests
+        uses: cypress-io/github-action@v2
+        with:
+          start: node ./scripts/createApp foo --acceptDefaults
+          spec: test/prisma.js
+        env:
+          NODE_ENV: test
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Run Cypress E2E Tests
         uses: cypress-io/github-action@v2
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,14 +31,12 @@ jobs:
       - name: Install packages
         run: |
           yarn install --frozen-lockfile
-      - name: Run Unit Tests
-        uses: cypress-io/github-action@v2
+      - name: Run Prisma Unit Tests
+        uses: actions/setup-node@v2.1.1
         with:
-          start: node ./scripts/createApp foo --acceptDefaults
-          spec: test/prisma.js
-        env:
-          NODE_ENV: test
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          node-version: 14.6.0
+        run: |
+          yarn test:prisma
       - name: Run Cypress E2E Tests
         uses: cypress-io/github-action@v2
         with:
@@ -49,7 +47,6 @@ jobs:
         env:
           NODE_ENV: test
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,9 +31,9 @@ jobs:
       - name: Install packages
         run: |
           yarn install --frozen-lockfile
-      - name: Run Prisma Unit Tests
+      - name: Run Unit Tests
         run: |
-          yarn test:prisma
+          yarn test:unit
       - name: Run Cypress E2E Tests
         uses: cypress-io/github-action@v2
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,9 +32,6 @@ jobs:
         run: |
           yarn install --frozen-lockfile
       - name: Run Prisma Unit Tests
-        uses: actions/setup-node@v2.1.1
-        with:
-          node-version: 14.6.0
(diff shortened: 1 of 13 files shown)


scripts/createAppAndStartServer.js
```diff
@@ -1,19 +1,30 @@
...
+const fs = require("fs");
 const os = require("os");
 const path = require("path");
-const mkdtemp = promisify(require("fs").mkdtemp);
+const mkdtemp = promisify(fs.mkdtemp);
 const execa = require("execa");
 module.exports = {};
 
-async function createAppAndStartServer() {
+async function init() {
```
alvincrespo, line 9: Just keeping the main method calls in sync across scripts.

scripts/createAppAndStartServer.js
```diff
@@ -1,19 +1,30 @@
...
   const appDir = await createApp(args);
   return startServer(appDir);
 }
 
+async function createTmpDir() {
+  const tmpDir = await mkdtemp(`${os.tmpdir()}${path.sep}`);
+
+  fs.writeFile("./tmp/tmpDir", tmpDir, () => {
+    console.log("tmpDir location stored.");
+  });
```
alvincrespo, line 20: We're storing the tmp directory location so we can pick it up easily if needed.

template/package.json.ejs
```diff
@@ -11,6 +11,7 @@
     "cypress:local": "CYPRESS_LOCAL=true CYPRESS_BASE_URL=http://localhost:3000 cypress",
     "db:migrate": "yarn -s prisma migrate up --experimental",
     "db:setup": "yarn db:migrate && yarn prisma generate",
+    "db:drop": "source prisma/.env && dropdb $(echo $DATABASE_NAME)",
```
alvincrespo, line 14: `dropdb` is a command line helper for postgres. Comes with every install - more [here](https://www.postgresql.org/docs/9.3/app-dropdb.html).

alvincrespo:
@cball I wasn't sure if we wanted to use Mocha for unit tests - but went down that route to at least get something running. Let me know if you have a library pref and I can update.

template/package.json.ejs
cball: what do you think about putting this into `scripts/dropDatabase`? If we did that we could do:

```js
import { Client } from 'pg';

const connectionString = process.env.DATABASE_URL;
const db = new Client({ connectionString });
const { database } = db.connectionParameters;

async function drop() {
  await db.connect();
  console.log(`dropping the ${database} database`);
  await db.query(`DROP DATABASE IF EXISTS ${database}`);

  await db.end();
}

drop();
```

cball: the big advantage here I think is that we can just use the existing `DATABASE_URL` and not have to manage another variable. we should have `pg` included as a dependency already.

.github/workflows/release.yml
```diff
@@ -31,6 +31,9 @@ jobs:
       - name: Install packages
         run: |
           yarn install --frozen-lockfile
+      - name: Run Prisma Unit Tests
+        run: |
+          yarn test:prisma
```
cball, line 36: should we make this generic unit tests?

package.json
```diff
@@ -44,6 +45,7 @@
     "@semantic-release/git": "^9.0.0",
     "cypress": "^5.1.0",
     "ejs-lint": "^1.1.0",
+    "mocha": "^8.1.3",
```
cball, line 48: Like I mentioned in the comment I think we should keep this as `jest` for continuity sake. glad you got this in though!

scripts/createApp.js
```diff
@@ -0,0 +1,47 @@
...
+
+  await execa("node", [cliPath, name, ...cliOptions], {
+    cwd: tmpdir,
+    stdio: "inherit",
+  });
+
+  return path.join(tmpdir, name);
+}
+
+init();
```
cball, line 47: 👍

template/prisma/_.env.ejs
```diff
@@ -5,3 +5,4 @@
 # See the documentation for all the connection string options: https://pris.ly/d/connection-strings
 
 DATABASE_URL="postgresql://<%= db.dev.user %><%= db.dev.password %>@<%= db.dev.host %>:<%= db.dev.port %>/<%= db.dev.name %>?schema=public"
+DATABASE_NAME="<%= db.dev.name %>"
```
cball, line 8: if we go the scripts route, may not need this

cball (changes requested):
awesome @alvincrespo! It might be good to stick with `jest` since that's what we are leveraging once the app is generated. Looking at what you have for Mocha I don't think it will be much of an update.

Along the lines of what you mentioned, I'd be curious if our unit tests need to go through the whole app generation process, but we can circle back to that.

The biggest suggestion I have is around creating a script to drop the db. If we go that route, you'll want something like this:

```json
{
  "db:drop": "DOTENV_CONFIG_PATH=./prisma/.env ts-node -r dotenv/config ./scripts/dropdb",
}
```

alvincrespo:
@cball Awesome recommendations! This makes sense to me. I'll go ahead and implement.

alvincrespo:
@cball Ready for another round of review.

cball (approved):

PR #79, opened 2020-10-10
fix: ensure @nexus/schema config uses prettier config

### LINKED ISSUE #77: Nexus schema generation should run prettier before write
Discovered when looking at https://github.com/echobind/bisonapp/issues/76.

image

cball:
This ensures our generated schema runs through prettier. This also applies the graphql best practice of alphabetically sorting field names in the schema file.

## Changes

- Add prettier config to nexus schema config
- Run prettier on `api.graphql` and and prettier.config
- Add graphql files to `lint-staged` as a fallback

## Screenshots

n/a

## Checklist

Fixes #77


template/api.graphql
```diff
@@ -1,60 +1,91 @@
...
+  gt: DateTime
+  gte: DateTime
+  in: [DateTime]
+  lt: DateTime
+  lte: DateTime
+  not: NestedDateTimeFilter
+  notIn: [DateTime]
+}
+
+input NestedStringFilter {
```
isaiahgrey93, line 65: Curious to why this exists being that there doesn't appear to be any difference from `StringFilter` as far as I can tell?
Maybe just semantics in the name?

Same thought on the other types with the `Nested{Type}` 🤔

template/package.json.ejs
```diff
@@ -114,6 +114,7 @@
     }
   },
   "lint-staged": {
-    "*.{ts,tsx}": "yarn lint"
+    "*.{ts,tsx}": "yarn lint",
+    "*.{graphql}": "prettier --write"
```
isaiahgrey93, line 118: ✨

isaiahgrey93 (approved):

template/api.graphql
cball: Good question. It looks like we forgot to re-run this during an upgrade. After all the open PRs are merged, I'll re-run and check the diff.

PR #80, opened 2020-10-10
fix:  lint error in dropDatabase broke codegen

### LINKED ISSUE #78: Linting error is causing codegen to fail
Discovered when looking at https://github.com/echobind/bisonapp/issues/76.

The newly added `dropDatabase` script had a linting error. That should be addressed and we should add lint as a CI step.

cball:
There was a linting error in the `dropDatabase` script that was causing codegen to fail.

## Changes

- disable eslint in the script

## Screenshots

Before:
CleanShot 2020-10-10 at 11 29 44@2x

After:
CleanShot 2020-10-10 at 12 17 39@2x

## Checklist

Fixes #78

### DIFF
--- a/template/scripts/dropDatabase.ts
+++ b/template/scripts/dropDatabase.ts
@@ -1,6 +1,8 @@
 export {};
 
+// eslint-disable-next-line
 const url = require('url');
+
 const Client = require('pg').Client;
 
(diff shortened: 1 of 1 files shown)


isaiahgrey93 (approved):
🧹

PR #81, opened 2020-10-11
fix: remove whitespace from package.json

cball:
This was outputting blank lines in the package.json, which we don't want. this PR fixes it.

## Changes

- Clean up `package.json` template

## Screenshots
n/a

## Checklist


isaiahgrey93 (approved):
:slaps_roof_of_car:

Curious if there is any prettier support for .ejs?

cball:
Not that I know of. Especially since whitespace is significant.

One thing we could do is run prettier on the outputted file and make sure there are no changes?

Issue #66, opened 2020-10-01
Migrate back to nexus/schema from nexus/framework


cball:
See: https://github.com/prisma-labs/graphql-framework-experiment/issues/1432.

I'm going to follow the migration guide they posted here: https://nexusjs.org/docs/adoption-guides/neuxs-framework-prisma-users

PR #93, opened 2020-10-20
fix(tests): ensure the generated tests pass

### LINKED ISSUE #92: Generated tests are failing after fresh install
This is likely an artifact from #68. At some point we need a CI step that runs tests in the generated app.

cball:
This fixes a few issues from #66 that were missed and caused tests in a newly generated app to fail.

## Changes

- regenerate `api.graphql` file
- adds `prisma` alias to context
- relaxes regex around test schema parsing to avoid database truncation errors
- updates snapshots to reflect error message and cache control updates

## Screenshots
image

## Checklist

Fixes #92

### DIFF
--- a/template/api.graphql
+++ b/template/api.graphql
@@ -25,11 +25,15 @@ input DateTimeFilter {
   equals: DateTime
   gt: DateTime
   gte: DateTime
-  in: [DateTime]
+  in: [DateTime!]
   lt: DateTime
   lte: DateTime
   not: NestedDateTimeFilter
-  notIn: [DateTime]
+  notIn: [DateTime!]
+}
+
+input EnumRoleNullableListFilter {
+  equals: [Role!]
 }
 
 """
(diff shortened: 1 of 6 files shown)


template/api.graphql
```diff
@@ -134,11 +138,16 @@ type Query {
...
+    orderBy: [UserOrderByInput!]
     where: UserWhereInput
   ): [User!]!
 }
 
+enum QueryMode {
+  default
+  insensitive
+}
+
```
cmejet, line 150: Nice enum!

template/tests/helpers.ts
```diff
@@ -53,7 +53,7 @@ export const graphQLRequestAsUser = async (
 export const resetDB = async (): Promise<boolean> => {
   if (process.env.NODE_ENV === 'production') return Promise.resolve(false);
 
-  const match = process.env.DATABASE_URL.match(/schema=(\w*)(&.*)*$/);
+  const match = process.env.DATABASE_URL.match(/schema=(.*)(&.*)*$/);
```
cmejet, line 56: Should we have a function with a test for this to validate the regex?

cmejet (approved):

PR #68, opened 2020-10-06
feat: migrate to @nexus/schema from nexus framework

cball:
For background, see: https://github.com/prisma-labs/graphql-framework-experiment/issues/1432

## Changes

- Add a tsconfig.cjs.json so that scripts can use `import` instead of `require`
- Remove explicit generate of Nexus Types
- Explicitly add prisma to context object
- Add `graphql-scalars`, which were included with Nexus Framework
- Remove redundant `graphql-iso-date` scalar
- Move to `apollo-server-micro` if not on Heroku
- add `copyDirectoryWithTemplate` function
- supports both `apollo-server-micro` (for Vercel) and `apollo-server-express` (for Heroku)

## Screenshots

n/a

## Checklist

- [x] Requires dependency update?
- [x] Generating a new app works

Fixes #66

BREAKING CHANGES:
* The project now depends on `@nexus/schema` instead of Nexus Framework.
* Move graphql files into `graphql/modules` for better organization and easier import to schema

PR #95, opened 2020-10-22
feat: add migrate on deploy and bison config to package.json

cball:
This ensures that builds are migrated properly for branches we care about (the staging and prod branches)

In order to properly do this, we add a bison config to package.json. That stores:
* The version used to init the project
* the staging branch
* the production branch

Once we move to Lerna and have a separate Bison package, we wont need the version. But until that time it will be really helpful in helping users upgrade to the latest improvements.

## Changes

- add `buildProd` script which handles migrate & build
- ask users what branches to use for staging/production. This is to drive migrations for staging/prod databases, but also paves the way in the future to help handle branch merging or other branch based configurations.
- forces `build:nexus` to use `NODE_ENV=development`. This is a huge gotcha, but if `NODE_ENV` is set to production, it will not build types and thus will fail on CI.

## Screenshots
image

## Checklist

### DIFF
--- a/template/.github/workflows/main.js.yml.ejs
+++ b/template/.github/workflows/main.js.yml.ejs
@@ -118,9 +118,9 @@ jobs:
     ## For a traditional flow that auto-deploys staging and deploys prod is as needed, keep as is
   
 <% if (host.name === 'heroku') { -%>
-    if: github.ref == 'refs/heads/staging' 
+    if: github.ref == 'refs/heads/<%= repo.stagingBranch %>' 
 <% } else { -%>
-    if: github.ref != 'refs/heads/production' # every branch EXCEPT production
+    if: github.ref != 'refs/heads/<%= repo.productionBranch %>' # every branch EXCEPT production
 <% } -%>
     steps:
       - uses: actions/checkout@v2
@@ -148,7 +148,7 @@ jobs:
     needs: tests
     ## For a typical JAMstack flow, the ref below should be your default branch.
     ## For a traditional flow that auto-deploys staging and deploys prod is as needed, keep as is
-    if: github.ref == 'refs/heads/production'
+    if: github.ref == 'refs/heads/<%= repo.productionBranch %>'
     steps:
       - uses: actions/checkout@v2
 <% if (host.name === 'vercel') { -%>
--- a/template/package.json.ejs
+++ b/template/package.json.ejs
@@ -6,10 +6,10 @@
   "cacheDirectories": [".next/cache"],
 <% } -%>
   "scripts": {
-    "build": "yarn build:nexus && yarn build:prisma && yarn build:next",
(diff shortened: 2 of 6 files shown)


template/package.json.ejs
```diff
@@ -6,10 +6,10 @@
   "cacheDirectories": [".next/cache"],
 <% } -%>
   "scripts": {
-    "build": "yarn build:nexus && yarn build:prisma && yarn build:next",
+    "build": "ts-node ./scripts/buildProd",
     "build:prisma": "prisma generate",
     "build:next": "next build",
-    "build:nexus": "ts-node -P tsconfig.cjs.json --transpile-only graphql/schema.ts",
+    "build:nexus": "NODE_ENV=development ts-node -P tsconfig.cjs.json --transpile-only graphql/schema.ts",
```
cball, line 12: This is one of the biggest gotchas I've ever seen. Nexus will not generate types if `NODE_ENV=production`. It's not documented well, but I did eventually find one place where it was:

image

This bug came about after #68

template/scripts/buildProd.ts
```diff
@@ -0,0 +1,44 @@
...
+  const branchesToMigrate = new RegExp(`${staging}|${production}`);
+  const currentBranch = process.env.BRANCH || process.env.VERCEL_GITHUB_COMMIT_REF;
+  const shouldMigrate = branchesToMigrate.test(currentBranch);
+
+  console.log('--------------------------------------------------------------');
+  console.log('Determining if we should migrate the database...');
+  console.log('branches to migrate:', branchesToMigrate);
+  console.log('current branch name:', currentBranch || '(unknown)');
+  console.log(shouldMigrate ? `${currentBranch} detected. Migrating.` : `Not running migrations.`);
+  console.log('--------------------------------------------------------------');
```
cball, line 22: Users will see this log in the build, so it should be really clear. This isn't supported on Heroku, so we'll have to add the way to do it for that platform at some point.

cmejet: 👍

template/.github/workflows/main.js.yml.ejs
```diff
@@ -118,9 +118,9 @@ jobs:
     ## For a traditional flow that auto-deploys staging and deploys prod is as needed, keep as is
   
 <% if (host.name === 'heroku') { -%>
-    if: github.ref == 'refs/heads/staging' 
+    if: github.ref == 'refs/heads/<%= repo.stagingBranch %>'
```
cmejet, line 121: This is a great add for flexibility!

template/package.json.ejs
cmejet: I'm so glad you figured this one out! A great reason for us to be running the tests of the `created app`

cmejet (approved):

PR #96, opened 2020-10-22
chore: remove extra ts-node dependency

cball:
It is already included via ts-node-dev.

## Changes

- removes ts-node from package.json template.

## Screenshots

n/a

## Checklist

### DIFF
--- a/template/package.json.ejs
+++ b/template/package.json.ejs
@@ -104,7 +104,6 @@
     "start-server-and-test": "^1.11.2",
     "supertest": "^4.0.2",
     "ts-jest": "^26.1.3",
-    "ts-node": "^8.10.2",
     "ts-node-dev": "^1.0.0-pre.63",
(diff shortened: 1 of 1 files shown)


cmejet (approved):
Do we need to be worried about the version of ts-node-dev being `pre`?

template/package.json.ejs
```diff
@@ -104,7 +104,6 @@
     "start-server-and-test": "^1.11.2",
     "supertest": "^4.0.2",
     "ts-jest": "^26.1.3",
-    "ts-node": "^8.10.2",
     "ts-node-dev": "^1.0.0-pre.63",
```
cball, line 107:
```suggestion
    "ts-node-dev": "^1.0.0",
```

cball:
@cmejet not anymore, apparently they released 1.0 today!

PR #102, opened 2020-10-27
feat(template): e2e test / factory improvements, update chakra dependencies

cball:
This includes various improvements for tests, factories, and the cypress typescript setup.

## Changes

Template:
- add `buildPrismaIncludeFromAttrs` test helper. Used in factories to auto-build associations
- source `.env.local` if running Cypress via `CYPRESS_LOCAL`
- add `login` and `createUserAndLogin` Cypress commands
- convert Cypress command fiile to typescript
- fix tsconfig for cypress
- ensure we re-generate Prisma client on `db:migrate` and `db:rollback`
- add `test:e2e:local` script for running tests against local dev server
- add logout e2e test that uses `createUserAndLogin` test
- add missing `framer-motion` dependency
- update chakra dependencies

## Screenshots

n/a

## Checklist


template/cypress/support/commands.ts
```diff
@@ -0,0 +1,56 @@
...
+declare global {
+  // eslint-disable-next-line
+  namespace Cypress {
+    interface Chainable {
+      login: typeof login;
+    }
+  }
+}
+
+// -- This is a parent command --
```
cmejet, line 17: Are we keeping this here for information or should it be removed?

cmejet (approved):
LGTM just one question about dead code :)

template/cypress/support/commands.ts
cball: it's generated by default when you install Cypress. We can definitely take it out if you think it's not helpful.

PR #111, opened 2020-11-17
fix: setup node action

cmejet:
This PR bumps `actions/setup-node` to version `2.1.2` to address the deprecated `set-env` capability with git actions.

## Changes

- Bump version actions/setup-node in the template for main.yml workflow
- Bump version for tests workflow same as above

## Checklist


cmejet:
Worked off main instead of canary :(

PR #113, opened 2020-12-11
docs: Change all uses of JAMstack to Jamstack

ajcwebdev:
## Changes

JAMstack -> Jamstack

## Screenshots

Screen Shot 2020-12-11 at 4 34 45 PM

## Checklist

## Fixes

In [December of last year](https://github.com/jamstack/jamstack.org/commit/41c0b767694c1f8c7e3fabcb1e0d770b154c00d7) Netlify made a slight modification to the Jamstack brand that they have cultivated and removed the all uppercase JAM.

Jamstack_Logo_DarkBG_Solid

Here's Chris Coyier's take on it: [JAMstack vs. Jamstack](https://css-tricks.com/jamstack-vs-jamstack/). For me, the reasoning I would switch this Bison repo specifically doesn't really have anything to do with the argument about the acronym.

It's more so that I just think it's been long enough to definitively say it's not gonna change back. But if you want to keep it the way it is and go old school then more power to you! And I'd also change Full Stack to Fullstack, just cause Fullstack Jamstack looks so much cleaner to me. But that's just me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center" style="text-align:center">
   <img alt="Bison Logo" src="https://user-images.githubusercontent.com/14339/89243835-f47e7c80-d5d2-11ea-8d8d-36202227d0ec.png" />
-  <h1 align="center">The Full Stack JAMstack in-a-box.</h1>
+  <h1 align="center">The Full Stack Jamstack in-a-box.</h1>
 </p>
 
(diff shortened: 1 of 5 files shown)


cball (approved):
Totally agree! Can you tell we've been operating in this ecosystem for a while? 😂

Appreciate the update!

ajcwebdev:
Good deal, we're a fast moving bunch here so you gotta keep up! And the last one is in the About section; you can't do a pull request for that.

Screen Shot 2020-12-11 at 7 18 32 PM

PR #112, opened 2020-11-18
feat: git action & chakra-ui update

cmejet:
This PR updates the git action for node-setup that addresses the deprecated set-env commands. It also updates Chakra to the latest stable 1.0 release 

## Changes

- Update git workflow actions for node setup
- Update chakra UI and associated templates/files

## Screenshots

(prefer animated gif)

## Checklist

- [ ] Requires dependency update?
- [X] Generating a new app works

PR #117, opened 2020-12-13
ci: update GH Action & Cypress dependencies for a successful build.

### LINKED ISSUE #116: GitHub Actions need to be updated (likely in template repo as well)
CleanShot 2020-12-11 at 21 19 54

cball:
There were three major changes required here. The `setup-node` step was done by @cmejet in #112. I split it out to a separate PR so that we can keep the Chakra UI upgrade isolated.

This PR also replaces the forked version of `vercel-action` that we were using previously(https://github.com/ngduc/vercel-deploy-action) with the original one (https://github.com/amondnet/vercel-action). It has the required updates for the recent `set-env` deprecation from GitHub (https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

Finally, `@testing-library/cypress` combined with our main Cypress version was causing us to download it twice, then fail the build on a checksum error. For more background, see: https://github.com/cypress-io/cypress/issues/4595

## Changes

- update `setup-node` to v2.1.2
- update `vercel-action` to v20
- update `@testing-library/cypress` to `^7.0.2`
- update `cypress` to `^6.1.0`

## Screenshots

CleanShot 2020-12-13 at 11 45 42

## Checklist

- [x] Tests in the generated app work

Fixes #116


packages/create-bison-app/template/.github/workflows/main.js.yml.ejs
```diff
@@ -125,14 +125,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 <% if (host.name === 'vercel') { -%>
-      - uses: ngduc/vercel-deploy-action@master
+      - uses: amondnet/vercel-action@v20
```
mcavaliere, line 128: 💯

packages/create-bison-app/template/.github/workflows/main.js.yml.ejs
```diff
@@ -125,14 +125,14 @@ jobs:
...
         with:
           vercel-token: ${{ secrets.VERCEL_TOKEN }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
           vercel-org-id: ${{ secrets.VERCEL_ORG_ID}}
           vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID}}
           # SET THIS OR YOU WILL DEPLOY TO YOUR PERSONAL VERCEL NAMESPACE
-          # scope: ${{ secrets.VERCEL_TEAM_ID }}
+          # scope: ${{ secrets.VERCEL_ORG_ID }}
```
mcavaliere, line 135: Just for background, how will users know to uncomment this? Is it in the tutorial?

mcavaliere (approved):
Looks good to me, FWIW.

packages/create-bison-app/template/.github/workflows/main.js.yml.ejs
cball: It could definitely use some more documentation. I think the correct answer here is to prompt for this value when creating the project. I'll add a followup issue for that.

Issue #47, opened 2020-09-02
reword database cli prompts for clarity


cball:
![image](https://user-images.githubusercontent.com/14339/91992790-8c6ea380-ed02-11ea-95e2-95efb3e52106.png)

mtliendo:
blog post link: https://lelon.io/blog/getting-started-with-bison

mtliendo:
Thoughts on enhancing the DX by having the second question ask _if_ they have a db setup?

If yes, proceed with the following questions. If no, walk them through steps similar to `yarn db:setup`?

cball:
@mtliendo I missed your comment here, sorry! I think that's a good plan. Right now we also assume postgres, so we may want to pave the way for updating that in the future.

Maybe the flow could be:

- What type of database would you like to use? (multiple choice, postgres only for now)
- Do you have it setup already? (y/n)
- if yes, ask above questions but reword for clarity
- if no, guide them through it? (brew install postgres && createuser postgres --superuser)

If the last step is too much to deal with, we could always link them to a markdown file in the repo.

Totally open to suggestions!

deloristhompson:
[PR123](https://github.com/echobind/bisonapp/pull/123)

PR #123, opened 2020-12-31
fix: Reword database cli prompts for clarity

deloristhompson:
Reword database cli prompts

## Changes

- Adds a prefix to each prompt

## Screenshots
Screen Shot 2020-12-31 at 10 43 14 AM

## Checklist
n/a

Fixes [Issue47](https://github.com/echobind/bisonapp/issues/47)


packages/create-bison-app/cli.js
```diff
@@ -54,35 +55,40 @@ function generateQuestions(appName) {
     {
       name: "db.dev.name",
       type: "input",
-      message: "What is the local database name?",
+      prefix: "The following link <link> will help you setup a new Postgres database if needed.\n\bThe name you will use for the new database (or current).\n",
```
cball, line 58: What do you think about asking if they have Postgres setup first?

> Do you have Postgres setup? y/n

(If the user answers no)
> Visit https://github.com/echobind/bisonapp/blob/main/docs/postgres.md for instructions!

(If the user answers yes, prompt for the database info below).

I like that you added both Homebrew and Postgres.app docs. Could we create a parent `Postgres` doc that links to each of those? That way, we can point users to a single place and they can choose how to proceed.

docs/homebrewPostgresInstructions.md
```diff
@@ -0,0 +1,12 @@
+# Homebrew - (Mac Only)
+## Installing PostgreSQL
+
+1. Install using hombrew: `brew install postgres`
+2. Add a local user (we recommend creating as a superuser to keep setup simple):  `create user postgres --superuser`
```
mcavaliere, line 5: Doesn't the user called `postgres` exist by default? Either way I recommend calling it something different. 

It's worth also recommending use of Postgres.app, which I find way easier to work with. 


https://postgresapp.com/

docs/postgresAppInstructions.md
```diff
@@ -0,0 +1,31 @@
+# Postgres.app (Mac Only)
```
mcavaliere, line 1: These are great!

docs/postgresqlEDBInstructions.md
```diff
@@ -0,0 +1,29 @@
...
+[PostgreSQL Database Download](https://www.enterprisedb.com/downloads/postgres-postgresql-downloads)
+
+## What is EDB Postgres?
+
+> A scalable platform designed to help users with everything from maintenance and distribution to support and documentation - *EDB.com*
+
+
+## Operating Systems:
+
+macOS
```
mcavaliere, line 11: Should this be a bulleted list maybe?

packages/create-bison-app/cli.js
```diff
@@ -51,38 +52,58 @@ function generateQuestions(appName) {
...
+      message: `${chalk.green("\b?")} Local database host`,
       description: "The database host",
       default: "localhost",
     },
     {
       name: "db.dev.port",
       type: "input",
-      message: "What is the local database port?",
+      prefix: "The port you will use for the new database (or current).\n",
+      message: `${chalk.green("\b?")} Local database port`,
```
mcavaliere, line 106: 👍 looking good

cball:
We're going to be revisiting the CLI prompts / questions.

PR #130, opened 2021-03-23
feat: upgrade prisma to 2.19.0

cmejet:
This PR upgrades Prisma to take advantage of the latest migrations and features available.

## Changes

- Resets the Migrations to the new SQL format
- Updates the the schema to match the new nexus-prisma-plugin format
- Updates the db context accessors to utilize the updates `findUnique` instead of `findOne`

## Screenshots

(prefer animated gif)

## Checklist

Fixes #xxx


mthomps4 (approved):

packages/create-bison-app/template/package.json.ejs
```diff
@@ -12,8 +12,8 @@
     "build:nexus": "NODE_ENV=development ts-node -P tsconfig.cjs.json --transpile-only graphql/schema.ts",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "db:migrate": "yarn -s prisma migrate up --experimental && yarn build:prisma",
-    "db:rollback": "yarn prisma migrate down --experimental && yarn build:prisma",
+    "db:migrate": "yarn -s prisma migrate deploy && yarn build:prisma",
+    "db:rollback": "yarn prisma migrate reset && yarn build:prisma",
```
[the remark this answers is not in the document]
cmejet: That is great feedback! I think we should rename it. It does actually do a full reset and replays migrations. I need to dig back through but there wasn't a rollback per say, I think you literally need to reverse the SQL migration or do via a migration.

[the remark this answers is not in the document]
mthomps4: @cmejet 

Weird... that seems like an oversight on their part. 
--
Actually just realized the new SQL migrations aren't in an `up`/`down` flow. 
Maybe they add that in the next run... 🙏🏼 

-- 

I would def rename this to `reset` if that is the case. 
We can figure out what a true `rollback` looks like later. 👍🏼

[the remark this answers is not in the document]
cmejet: A good article I found as well about this specifically. I also like that they made a change that when you call reset it will run the seed file if it detects it there. 

https://www.prisma.io/docs/concepts/components/prisma-migrate/prisma-migrate-limitations-issues#lack-of-rollbacks--down-migrations

packages/create-bison-app/template/graphql/context.ts
```diff
@@ -16,7 +16,7 @@ export async function createContext(context: ApolloApiContext): Promise<Context>
   let user: User | null = null;
 
   if (authHeader) {
-    user = await prisma.user.findOne({ where: { id: authHeader.userId } });
+    user = await prisma.user.findUnique({ where: { id: authHeader.userId } });
```
cmejet, line 19: This is a syntax change with the newest Prisma

packages/create-bison-app/template/graphql/modules/user.ts.ejs
```diff
@@ -178,6 +178,6 @@ export const SignupInput = inputObjectType({
   definition: (t) => {
     t.string('email', { required: true });
     t.string('password', { required: true });
-    t.field('profile', { type: 'ProfileCreateOneWithoutUserInput' });
+    t.field('profile', { type: 'ProfileCreateNestedOneWithoutUserInput' });
```
cmejet, line 181: Another syntax change with latest Prisma, discovered through build error and tracking down type generations

packages/create-bison-app/template/graphql/schema.ts
```diff
@@ -13,6 +13,7 @@ export const schema = makeSchema({
   types,
   plugins: [
     fieldAuthorizePlugin(),
+    declarativeWrappingPlugin(),
```
cmejet, line 16: This was required to add to the configuration in order to satisfy some of the current schema definitions we had.
https://nexusjs.org/docs/plugins/declarativeWrapping

packages/create-bison-app/template/graphql/schema.ts
```diff
@@ -27,18 +28,17 @@ export const schema = makeSchema({
...
+  sourceTypes: {
+    modules: [
       {
-        source: path.join(currentDirectory, 'node_modules/.prisma/client/index.d.ts'),
+        module: path.join(currentDirectory, 'node_modules/.prisma/client/index.d.ts'),
         alias: 'db',
       },
       {
-        source: path.join(currentDirectory, 'graphql', 'context.ts'),
+        module: path.join(currentDirectory, 'graphql', 'context.ts'),
```
cmejet, line 38: This is the updated config Syntax, not documented well for the migration

packages/create-bison-app/template/package.json.ejs
```diff
@@ -6,14 +6,15 @@
   "cacheDirectories": [".next/cache"],
 <% } -%>
   "scripts": {
-    "build": "ts-node ./scripts/buildProd",
+    "build": "node ./scripts/buildProd",
```
cmejet, line 9: Updated to a js file due to build errors that started occurring. The script did not appear to need to be typescript as it was already using exports.module syntax and an empty export.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -6,14 +6,15 @@
...
     "build:prisma": "prisma generate",
     "build:next": "next build",
     "build:nexus": "NODE_ENV=development ts-node -P tsconfig.cjs.json --transpile-only graphql/schema.ts",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "db:migrate": "yarn -s prisma migrate up --experimental && yarn build:prisma",
-    "db:rollback": "yarn prisma migrate down --experimental && yarn build:prisma",
+    "db:migrate": "yarn -s prisma migrate dev && yarn build:prisma",
+    "db:deploy": "yarn -s primsa deploy && yarn build:prisma",
+    "db:reset": "yarn prisma migrate reset && yarn build:prisma",
```
cmejet, line 17: Prisma no longer offers a direct rollback feature or `migrate down` so we renamed the script to match the functionality which resets the database and then runs the seed file if found. This is intended to be used on dev only, not on production database.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -6,14 +6,15 @@
...
+    "build": "node ./scripts/buildProd",
     "build:prisma": "prisma generate",
     "build:next": "next build",
     "build:nexus": "NODE_ENV=development ts-node -P tsconfig.cjs.json --transpile-only graphql/schema.ts",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "db:migrate": "yarn -s prisma migrate up --experimental && yarn build:prisma",
-    "db:rollback": "yarn prisma migrate down --experimental && yarn build:prisma",
+    "db:migrate": "yarn -s prisma migrate dev && yarn build:prisma",
+    "db:deploy": "yarn -s primsa deploy && yarn build:prisma",
```
cmejet, line 16: Prisma no longer uses the `migrate up` command and instead favors ` deploy`

packages/create-bison-app/template/package.json.ejs
```diff
@@ -22,7 +23,7 @@
     "g:component": "hygen component new --name",
     "g:graphql": "hygen graphql new --name",
     "g:page": "hygen page new --name",
-    "g:migration": "yarn -s prisma migrate save --experimental",
+    "g:migration": "yarn -s prisma migrate dev",
```
cmejet, line 26: Updated syntax for prisma 2.19.0

packages/create-bison-app/template/package.json.ejs
```diff
@@ -104,11 +104,12 @@
     "nanoid": "^3.1.10",
     "pg": "^8.3.0",
     "prettier": "^2.0.5",
+    "prisma": "^2.19.0",
     "start-server-and-test": "^1.11.2",
     "supertest": "^4.0.2",
     "ts-jest": "^26.1.3",
     "ts-node-dev": "^1.0.0",
-    "typescript": "^3.9.7"
+    "typescript": "^4.1.3"
```
cmejet, line 112: Updated to match peer dependencies.

packages/create-bison-app/template/tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "forceConsistentCasingInFileNames": true,
     "noEmit": true,
     "esModuleInterop": true,
-    "module": "esnext",
+    "module": "commonjs",
```
cmejet, line 11: All other tsconfig files are using commonjs and this started creating a build error when running buildProd, it was not happy with the empty export.

packages/create-bison-app/template/package.json.ejs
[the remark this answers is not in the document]
mthomps4: After thinking about this more and chatting with Cory on Nest/CS work -- I get the reasoning... Still feels odd, but I get it.

packages/create-bison-app/template/package.json.ejs
mcavaliere: I fixed this in the CITJS repo by overriding the module definition in the CLI params: 

```"ts-node": "ts-node --compiler-options '{\"module\":\"CommonJS\"}'",```

Prisma has some notes on this:

https://www.prisma.io/docs/guides/application-lifecycle/seed-database#set-up-seeding-in-typescript

File: packages/create-bison-app/template/package.json.ejs [hunk 12]
```diff
@@ -62,28 +62,28 @@
...
   "devDependencies": {
     "@apollo/react-testing": "^4.0.0",
-    "@graphql-codegen/cli": "1.17.6",
-    "@graphql-codegen/typescript": "1.17.6",
+    "@graphql-codegen/cli": "^1.21.3",
+    "@graphql-codegen/typescript": "^1.17.6",
     "@graphql-codegen/typescript-operations": "^1.17.6",
-    "@graphql-codegen/typescript-react-apollo": "1.17.6",
-    "@graphql-codegen/typescript-resolvers": "1.17.4",
-    "@prisma/cli": "2.8.1",
```
mcavaliere, line 78: 👏

packages/create-bison-app/template/tsconfig.json
mcavaliere: See my comment above,  you can still set this to esnext and override at the command level

mcavaliere (changes requested):
Small comments - only blocking change I see is trying the commonjs override so we can keep using ts-node for the build script.

MariahGrey (approved):
LGTM 👍

mcavaliere (approved):
LGTM

PR #147, opened 2021-05-27
feat: add cross-env package to normalize env vars across platforms

samuelkilada:
## Changes

- Added cross-env package
- Updated scripts to use cross-env in package.json

## Checklist

- [x] Generating a new app works (cross-platform tests are needed)


cball (approved):
Looks great! Thanks for working through this.

PR #160, opened 2021-06-15
Remove generated files from source

### LINKED ISSUE #154: Remove types from source control, add step to setup to generate them
These have been causing merge problems on projects. To determine impact we looked at what needs to exist to properly run everything:

```
To summarize what I found with those generated files and running yarn dev: If nexus.d.ts and api.graphql are missing, the codgen that yarn dev runs will fail and it won't generate types.ts. Also, for some reason nexus.d.ts and api.graphql will not be generated either. If only types.ts is missing, yarn dev will regenerate it. So it looks like at least nexus.d.ts and api.graphql need to exist before running yarn dev
```

This means that when someone first sets up the project, a setup script needs to exist that runs at a minmum:
* yarn db:migrate (creates db and generates prisma client)
* yarn nexus:build (generates nexus.d.ts)
* yarn codegen (generates types.ts and api.graphql)

Once those files exist, the user should be able to run `yarn dev` without issue.

Tasks:
- [ ] remove those files from the template
- [ ] add entries to .gitignore
- [ ] add script to run setup commands

kgajera:
## Changes

- Removes all generated files from source and adds them to `.gitignore`
- Adds new `yarn generate` script to generate prisma client, nexus typings, and GraphQL typings
  -  This is an extra step required when first creating the project. I didn't see a good way to combine the `yarn db:setup` and this together since they do very different things. Let me know if you have any other ideas or want changes to this.
- Upgrades graphql-codegen dependencies
- Fixes lint errors in the template so graphql-codegen will successfully complete
- Fixes some of the tests. I noticed we have tests for GraphQL queries that don't exist in our API (for example, `users`), I've updated to skip those tests.
- I think our CI was having memory issues with running the `yarn generate && yarn build:next` command on multiple jobs so I had to limit to running one job at a time

## Checklist

Fixes #154


packages/create-bison-app/template/scripts/buildProd.ts
```diff
@@ -1,7 +1,7 @@
 export {};
 const spawn = require('child_process').spawn;
 
-const DEFAULT_BUILD_COMMAND = `yarn build:nexus && yarn build:prisma && yarn build:next`;
+const DEFAULT_BUILD_COMMAND = `yarn generate && yarn build:next`;
```
cmejet, line 4: Ahhh, this answers my question from discord about when running yarn dev :) this is great!

packages/create-bison-app/template/tests/requests/user/signup.test.ts
```diff
@@ -48,15 +58,20 @@ describe('User signup mutation', () => {
...
+        },
       };
 
       const response = await graphQLRequest({ query, variables });
       const errors = response.body.errors.map((e) => e.message);
 
       expect(errors).toMatchInlineSnapshot(`
         Array [
-          "Variable \\"$data\\" got invalid value { email: \\"hello@wee.net\\", password: \\"fake\\", roles: { set: [Array] } }; Field \\"roles\\" is not defined by type \\"SignupInput\\".",
+          "Variable \\"$data\\" got invalid value { email: \\"hello@wee.net\\", password: \\"fake\\", profile: { create: [Object] }, roles: { set: [Array] } }; Field \\"roles\\" is not defined by type \\"SignupInput\\".",
```
cmejet, line 74: Great update!

packages/create-bison-app/template/tests/requests/user/users.test.ts
```diff
@@ -6,7 +6,7 @@ import { UserFactory } from '../../factories/user';
 beforeEach(async () => resetDB());
 afterAll(async () => disconnect());
 
-describe('users query', () => {
+describe.skip('users query', () => {
```
cmejet, line 9: I will work on a few of these in the next PR, fixed in another app

cmejet (approved):

PR #162, opened 2021-06-16
chore: upgrade react-hook-form

### LINKED ISSUE #159: Upgrade react-hook-form
This is a major version update which does contain breaking changes.

kgajera:
This upgrade ours `react-hook-form` to the latest version.

## Changes

- There were a couple breaking changes: https://github.com/react-hook-form/react-hook-form/blob/master/CHANGELOG.md#changes-16
  -  `register` now has to spread as props
  -  `errors` alias has been removed and now has to be accessed through `formState`

## Checklist

Fixes #159

### DIFF
--- a/packages/create-bison-app/template/README.md.ejs
+++ b/packages/create-bison-app/template/README.md.ejs
@@ -237,7 +237,7 @@ export function OrganizationForm() {
 
   return (
     <form onSubmit={handleSubmit(onSubmit)}>
-      <input name="name" ref={register({ required: true })} />
+      <input {...register('name', { required: true })} />
       {errors.name && <span>This field is required</span>}
       
       <input type="submit" />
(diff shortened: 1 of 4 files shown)


packages/create-bison-app/template/README.md.ejs
```diff
@@ -237,7 +237,7 @@ export function OrganizationForm() {
 
   return (
     <form onSubmit={handleSubmit(onSubmit)}>
-      <input name="name" ref={register({ required: true })} />
+      <input {...register('name', { required: true })} />
```
cmejet, line 240: I like these changes with the upgrade, very readable!

cmejet (approved):
Generated new app and validated, works great! Same tests failing as previously when running yarn test from initialization, some resolved in #163

PR #168, opened 2021-06-22
feat: add dev server that will live reload on template changes

kgajera:
This improves the developer experience and productivity to contribute to the bison template. It enables you to  make changes directly to files in `packages/create-bison-app/template/` and have it live reload a running bison app.

This adds a `yarn dev` script in `packages/create-bison-app` which will do the following:

1. Creates a new bison app in `packages/create-bison-app/dist/bison-dev-app`
   -  If `packages/create-bison-app/dist/bison-dev-app` already exists, this step will be skipped
   -  A `bison.json` file will also be created that contains all the variables for template rendering. This is needed later to watch and re-render EJS files. This file will **only** be created for this dev app and not other bison apps.
   -  Symbolic links to node modules and generated files are created in `packages/create-bison-app/template/` so the files won't contain errors for missing imported modules
1. Runs `yarn db:setup && yarn generate`
   -  If `packages/create-bison-app/dist/bison-dev-app` already exists, this step will be skipped
1. Uses our `startServer` script to run `yarn next dev` 
1. Starts watching the `packages/create-bison-app/template/` folder for changes. Anytime a file is changed, it will be copied to `packages/create-bison-app/dist/bison-dev-app/` where the app is actually running.
   -  For EJS files, it will use the `bison.json` file mentioned above as variables to render the template

Examples of supported arguments:
- `yarn dev --acceptDefaults`
   -  Skip interactive prompts if a new bison app is being created
- `yarn dev --clean`
   -  This will delete `packages/create-bison-app/dist/bison-dev-app`, unlink node modules and generated files, and create a new bison app
- `yarn dev --port=5000`
    - Port to use for dev app. Defaults to 3001.

## Checklist

### DIFF
--- a/packages/create-bison-app/scripts/createDevAppAndStartServer.js
+++ b/packages/create-bison-app/scripts/createDevAppAndStartServer.js
@@ -1,0 +1,107 @@
+const chokidar = require("chokidar");
+const execa = require("execa");
+const fs = require("fs");
+const path = require("path");
+const yargs = require("yargs/yargs");
+const { createApp } = require("../scripts/createApp");
+const { startServer } = require("../scripts/startServer");
+const { templateFolder } = require("../tasks/copyFiles");
+const { cleanTemplateDestPath } = require("../utils/copyDirectoryWithTemplate");
+const { copyWithTemplate } = require("../utils/copyWithTemplate");
+
+const BISON_DEV_APP_NAME = "bison-dev-app";
+const BISON_DEV_APP_VARIABLES_FILE = "bison.json";
+
+async function init() {
+  const distPath = path.join(__dirname, "..", "dist");
+  const devAppPath = path.join(distPath, BISON_DEV_APP_NAME);
+  const args = process.argv.slice(2);
+  const { clean, port } = yargs(args).argv;
+
+  if (clean) {
+    await fs.promises.rmdir(devAppPath, { recursive: true });
+    await removeTemplateSymlinks();
+  }
+
+  // Create bison app if it does not exist
+  if (!fs.existsSync(devAppPath)) {
+    if (!fs.existsSync(distPath)) {
+      await fs.promises.mkdir(distPath);
+    }
+
+    await createApp([BISON_DEV_APP_NAME, ...args], distPath);
+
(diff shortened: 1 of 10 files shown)


cmejet (dismissed):
This is SLICK! I LOVE IT! Definitely makes contributing to bison much easier! I can save this would have saved me a TON of time recently and look forward to using it.

Only ask is can we please add the comments in this PR to a markdown file that documents this so we don't lose it?
Also to note was just a reminder to make sure you run `yarn` from within the create-bison-app folder before running `yarn dev`

packages/create-bison-app/utils/copyDirectoryWithTemplate.js
```diff
@@ -20,16 +30,13 @@ async function copyDirectoryWithTemplate(from, to, variables) {
 
   return await Promise.all(
     files.map(async (file) => {
-      const toFile = file
-        .replace(from, to)
-        .replace(/_\./, ".")
-        .replace(/\.ejs$/, "");
-
+      const toFile = cleanTemplateDestPath(file.replace(from, to));
       return copyWithTemplate(file, toFile, variables);
```
cball, line 34: 👍

packages/create-bison-app/scripts/createDevAppAndStartServer.js
```diff
@@ -0,0 +1,107 @@
...
+      await fs.promises.copyFile(src, dest);
+    }
+  };
+
+  // Watch template files and copy to dev app
+  chokidar.watch(templateFolder, {
+      ignored: path => path.includes("node_modules")
+    })
+    .on("add", copyFile)
+    .on("change", copyFile);
```
cball, line 60: 👍

cball (commented):
Great job! Agree with @cmejet comments of adding a doc with these commands. How about we add a `CONTRIBUTING.md` to the project root?

kgajera:
I've added the contributing guides, one at the root and one in the `create-bison-app` package. Let me know if you want any changes.

cball (approved):
Looks great @kgajera! 🎉

PR #171, opened 2021-06-25
fix: update setupDB command to not fail when run multiple times

### LINKED ISSUE #170: Fix setupDb command for tests to reset database
https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/template/tests/helpers.ts#L88

The current command does not work when run multiple times on a database. It should use this: `yarn prisma migrate reset --force`

kgajera:
This fixes the command being run in `setupDB` to not fail in cypress tests when run multiple times on the same database.

## Changes

- Replace `prisma deploy` with `prisma reset` so it will drop the database if it exists and then run migrations 

## Checklist

Fixes #170


cball (approved):
Awesome! We're not calling setupDB before every test are we? it should just be when the suite starts.

Either way this change is 👍

Issue #173, opened 2021-06-29
Fix generated GraphQL module template to create correct "where" inputs


kgajera:
"User" is hard-coded in the template: https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/template/_templates/graphql/new/graphql.ejs#L92

PR #174, opened 2021-06-29
feat(enhancement): Upgrades generated graphql nexus file with more CRUD examples

mthomps4:
## Description 
Upgrades generated graphql nexus file with more CRUD examples and proper Input Types

## Changes
GraphQL Nexus template was lacking, had bogus "SOMETHING" and User definitions. 

Now generates: 
- Base Module 
- Query `thing` w/ user logged in guard
- Query `list('thing')` w/ user logged in guard
- Mutation `createThing` w/ `isAdmin` guard
- Mutation `updateThing` w/ `IsAdmin` guard
- ThingWhereInput
- ThingWhereUniqueInput 
- ThingCreateInput
- ThingUpdateInput
- ThingOrderByInput

closes https://github.com/echobind/bisonapp/issues/173

### DIFF
--- a/packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
+++ b/packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
@@ -3,27 +3,18 @@ to: graphql/modules/<%= name %>.ts
 ---
 <% camelized = h.inflection.camelize(name) -%>
 <% plural = h.inflection.pluralize(camelized) -%>
-import { objectType, extendType, inputObjectType, list, arg, nonNull /*, enumType*/ } from 'nexus';
+import { objectType, extendType, inputObjectType, arg, nonNull /*, enumType*/ } from 'nexus';
 
 import { isAdmin } from '../../services/permissions';
-
-// Example ENUM Type
-// import { Role } from '@prisma/client';
-// export const UserRole = enumType({
-//   name: 'Role',
-//   members: Object.values(Role),
-// });
 
 // <%= camelized %> Type
 export const <%= camelized %> = objectType({
   name: '<%= camelized %>',
   description: 'A <%= camelized %>',
   definition(t) {
-    // Database Specific Fields
     t.nonNull.id('id');
     t.nonNull.date('createdAt');
     t.nonNull.date('updatedAt');
-    // Model Specific Fields
     t.nonNull.string('name');
   },
 });
@@ -33,8 +24,8 @@ export const <%= camelized %>Queries = extendType({
   type: 'Query',
   definition: (t) => {
     // List <%= plural %> Query
-    t.field('<%= plural.toLowerCase() %>', {
-      type: list('<%= camelized %>'),
+    t.list.field('<%= plural.toLowerCase() %>', {
+      type: '<%= camelized %>',
       authorize: (_root, _args, ctx) => !!ctx.user,
       args: {
         where: arg({ type: '<%= camelized %>WhereInput' }),
--- a/packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
+++ b/packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
@@ -3,20 +3,28 @@ to: graphql/modules/<%= name %>.ts
 ---
 <% camelized = h.inflection.camelize(name) -%>
 <% plural = h.inflection.pluralize(camelized) -%>
-import { objectType, extendType, inputObjectType, stringArg, arg, nonNull, enumType } from 'nexus';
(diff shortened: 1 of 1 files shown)


packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
```diff
@@ -3,20 +3,28 @@ to: graphql/modules/<%= name %>.ts
...
 
-// import { isAdmin } from '../services/permissions';
+import { isAdmin } from '../../services/permissions';
+
+// Example ENUM Type
+// import { Role } from '@prisma/client';
+// export const UserRole = enumType({
+//   name: 'Role',
+//   members: Object.values(Role),
+// });
```
cball, line 15: it feels like we might delete this most of the time. I definitely see the value in outlining the `Object.values` example, but I did want to ask if you think this falls into the "you'll need this often" category.

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
```diff
@@ -3,20 +3,28 @@ to: graphql/modules/<%= name %>.ts
...
 export const <%= camelized %> = objectType({
   name: '<%= camelized %>',
   description: 'A <%= camelized %>',
   definition(t) {
+    // Database Specific Fields
     t.nonNull.id('id');
     t.nonNull.date('createdAt');
     t.nonNull.date('updatedAt');
+    // Model Specific Fields
+    t.nonNull.string('name');
```
cball, line 27: I like the idea of trying to organize some of this. I worry that doing it this way may actually confuse things though. It requires you to ask yourself "is this a database field or a model field". Additionally, the `name` field that you have here doesn't have a custom resolver so it looks like it would be coming from the database.

Perhaps it would be better not to force the organization and just show an example of a custom resolver for a field?

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
```diff
@@ -25,21 +33,32 @@ export const <%= camelized %>Queries = extendType({
   type: 'Query',
   definition: (t) => {
     // List <%= plural %> Query
-    t.list.field('<%= plural.toLowerCase() %>', {
-      type: '<%= camelized %>',
+    t.field('<%= plural.toLowerCase() %>', {
+      type: list('<%= camelized %>'),
```
cball, line 37: correct me if I'm wrong but I think both of these are equivalent.  I feel `t.list.field` is slightly more straightforward due to the intellisense autocompletion and not requiring an additional import of the list function. thoughts?

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
```diff
@@ -25,21 +33,32 @@ export const <%= camelized %>Queries = extendType({
   type: 'Query',
```
cball, line 33: while you are here, thoughts on moving this to `queryField` instead of `extendType`? Not required, but it does make it a bit more concise. 

https://nexusjs.org/docs/api/query-field

cball (commented):
Looking great. Just a few minor questions

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
mthomps4: Agreed. 
It is not an everyday thing, probably something better suited for Docs over a generated file. 
Will opt to remove here.

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
mthomps4: Yup, those comments are a bit misleading. 
`name` is in our generated example. 

I'll remove the comments and leave it as is.

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
mthomps4: We made this change when troubleshooting the other day. 
It is the same... 

I'll revert and save an import. 👍🏼

packages/create-bison-app/template/_templates/graphql/new/graphql.ejs
mthomps4: Going to leave this one for another ticket. 
We'll want to refactor to use `query-field` and `mutation-field` throughout in our `user` module as well to stay consistent. 

I'll need to look at that syntax a bit, it looks like each piece would become its own exported `queryType`.

cball (approved):
Awesome! :shipit:

PR #178, opened 2021-06-30
fix: Removes generated id from factory in favor of DB

### LINKED ISSUE #144: remove `id` from factories in favor of database generation

mthomps4:
## Changes
fix: Removes generated id from factory in favor of DB

## Checklist

Fixes #144

### DIFF
--- a/packages/create-bison-app/template/_templates/test/factory/factory.ejs
+++ b/packages/create-bison-app/template/_templates/test/factory/factory.ejs
@@ -12,7 +12,6 @@ const chance = new Chance();
 export const <%= upper %>Factory = {
   build: (attrs: Partial<<%= upper %>> = {}) => {
     return {
-      id: chance.guid(),
       ...attrs,
     };
   },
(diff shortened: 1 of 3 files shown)


kgajera:
Should we also remove it from our `UserFactory`?

I think it will also cleanup this line and allow us to remove the eslint disable: https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/template/tests/requests/user/signup.test.ts#L94

mthomps4:
> Should we also remove it from our `UserFactory`?
> 
> I think it will also cleanup this line and allow us to remove the eslint disable: https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/template/tests/requests/user/signup.test.ts#L94

Yup, 
I'll make a note to clean that up. Forgot about the pre-existing pieces. 👍🏼

kgajera (commented):
Ran `yarn test` in a new bison app and the signup test is failing:

Screen Shot 2021-07-01 at 10 17 53 AM

mthomps4:
@kgajera Thanks for double-checking that! 
I removed `roles` previously from unused linting - and that was needing to be stripped for the test. 👍🏼 

---

Also, didn't realize UserFactory was still using the old custom Attrs instead of built types -- corrected that as well. 
```
Partial<Prisma.UserCreateInput> 
```


4fd8459

kgajera (approved):

PR #181, opened 2021-07-01
Heroku Revamp:

### LINKED ISSUE #125: createBisonApp function is called when user doesn't provide a git origin for Heroku
Message: `Would you like to add a git origin when complete? (required if using Heroku)`
All prompts for Heroku are bypassed due to the user selecting no to question.  Is this okay or should we add a validation where the user is prompted to provide the git URL after they select Heroku if they did not the first time since it is a requirement?

Screen Shot 2020-12-31 at 4 12 43 PM

### LINKED ISSUE #179: Heroku Deploy Docs
Instead of maintaining/integrating with Heroku's CLI. 
As a first pass opt to: 
- [ ] add a Procfile (and any other config docs) 
- [ ] add `docs/heroku-setup.md`
- [ ] If Heroku selected - Point the user to the generated markdown post-install. (docs/heroku-setup.md)

Note: this same pattern will be used for other deployment strategies such as `Render`, `Railway`, `Vercel` as to not make assumptions on deploy configurations. 

Closes #138 
Closes #139 
Closes #49

mthomps4:
## Changes
Heroku Revamp:
- Removes heroku specific questions from prompt
- Removes heroku CLI  specifics from main
- Adds Procfile to copied heroku files
- updates post-install text to point to heroku readme links
- updated conditional in readme for Deploy
(future placeholder for Render, Vercel, Other)

## Screenshots

image

## Checklist

Fixes #179 
Fixes #125


kgajera (commented):
Do we still need to display the "required if using Heroku" text for adding a git origin or can this be removed too: 
Screen Shot 2021-07-01 at 1 46 01 PM

Does this also close #125?

mthomps4:
@kgajera I was wondering about the order of these questions as well.
It feels like we would want to ask the "Hosting" question before this one.
And then nudge them on this after. 

cc @cball

cball:
If we are moving away from using the Heroku CLI than the branching for git+heroku is not required. I'd opt to remove all of it and just automatically commit for the user. they can handle setting up the remotes

mthomps4:
```
we do want to make the initial commit, just not add the origin
but there is a javascript only git package we can migrate to and remove the nodegit dependency
```

Removed "origin" logic and left the package swap for another new ticket.

packages/create-bison-app/template/.github/workflows/main.js.yml.ejs
```diff
@@ -100,71 +100,3 @@ jobs:
         with:
           name: cypress-videos
           path: cypress/videos
-
-  ###############
```
kgajera, line 104: Just want to double check that we really want to remove the Vercel deployment here too? I think that might be a good feature to keep in Bison if its working well. Setting up the CI can be time consuming so I think its really convenient when its provided out of the box.

If we do want to remove it, I think we'll have to remove the following lines from the main readme and also update the [CI](https://github.com/echobind/bisonapp/blob/GH179-Heroku-Setup-Docs/docs/ci.md) docs:



Screen Shot 2021-07-21 at 11 42 12 AM

mthomps4: @kgajera @cball -- From my last converstion, Chris mentioned taking it all out. 
To your point, if we are going to partner -- maybe that's the one flow we DO support. 

Probalby easier to find some five minutes for the three of us to chat. 
There is a lot of good code here.

kgajera (approved):

PR #186, opened 2021-07-12
feat(enhancement): generate placeholder graphql specs to match bolierplate

mthomps4:
Description of PR that completes issue here...

## Changes

- Generates a find, list, create, update spec to match generated graphql boilerplate. 

## Screenshots
generatedSpecs

## Checklist


kgajera (commented):
The test files look good. My only thought is if we should be adding these test files when running `g:graphql`. 

This is inconsistent with our other generate commands. It seems like we a have corresponding "test" generator for all our "new" generators (for example, `g:component` and `g:test:component`).

We have an existing `g:test:request` which generates this: https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/template/_templates/test/request/request.ejs

I think this PR adds an improved version of that. Does it make sense to delete the template linked above update `g:test:request` to generate the files added by this PR?

mthomps4:
@kgajera for sure -- This is def open for discussion w/ you and @cball 
It has helped us gain momentum pre-flight, so I wanted to throw it out here for debate.

mthomps4:
Update: 
Based on discussions, will be moving these to a Test Example File instead of auto generating. 
Leaving open as a reminder to self.

mthomps4:
Closed for #202 Documentation

PR #189, opened 2021-07-16
fix: refactor scripts import in the create-bison-app script into utils file

kgajera:
## Changes

The `scripts/*` files shouldn't be imported into other modules so this refactors getting and saving dev app variables into a utility file.

Fixes this error occurring when creating a new app:

Screen Shot 2021-07-15 at 5 08 01 PM

## Checklist

### DIFF
--- a/packages/create-bison-app/scripts/createDevAppAndStartServer.js
+++ b/packages/create-bison-app/scripts/createDevAppAndStartServer.js
@@ -8,9 +8,9 @@ const { startServer } = require("../scripts/startServer");
 const { templateFolder } = require("../tasks/copyFiles");
 const { cleanTemplateDestPath } = require("../utils/copyDirectoryWithTemplate");
 const { copyWithTemplate } = require("../utils/copyWithTemplate");
+const { getDevAppVariables } = require("../utils/devAppVariables");
 
 const BISON_DEV_APP_NAME = "bison-dev-app";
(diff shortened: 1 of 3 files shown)


packages/create-bison-app/scripts/createDevAppAndStartServer.js
```diff
@@ -112,6 +109,4 @@ if (require.main === module) {
   init();
 }
 
-module.exports = {
-  BISON_DEV_APP_VARIABLES_FILE,
-};
+module.exports = {};
```
nathanBrenner, line 112: Is this needed? It looks like you're exporting nothing from this module.

nathanBrenner (approved):
Nit: It would be nice if the pr template had a section on how to test the changes.

packages/create-bison-app/scripts/createDevAppAndStartServer.js
kgajera: Removed

kgajera:
> Nit: It would be nice if the pr template had a section on how to test the changes.

Agree with your suggestion.

You can test this by:
1. Run `npm link` in the `packages/create-bison-app` directory
2. Then you can create a new bison app, `npx create-bison-app MyApp`, and it will use your local project

cmejet (approved):
Working! Thanks for continuing to work on these.

PR #190, opened 2021-07-19
refactor: replace nodegit with isomorphic-git

### LINKED ISSUE #149: Replace nodegit with a more lightweight solution
nodegit takes a long time to install and often fails. We should look into a better solution since it's only used to create the initial commit. See what other projects are using for this.

kgajera:
## Changes

- Refactors usages of `nodegit` to `isomorphic-git`
   -  The only challenge was that there's no easy way to add all the untracked files. I followed their recommended approach: https://isomorphic-git.org/docs/en/faq#how-to-add-all-untracked-files-with-gitadd 
-  I noticed there were a few files being missed when copying from the template into a new bison project:
   -  `tsconfig.cjs.json` - This wasn't being used anywhere but I've updated the `yarn ts-node` script to use it. This should also fix running this in Windows because it didn't like the format of passing the compiler options in as arguments.
   -  `.nvmrc`
   - `cypress/.eslintrc.json` - this was not being copied because the `cpy` library excludes hidden files by default
-  Removes `packages/create-bison-app/yarn.lock`. I don't think this file was being used anymore after switching to lerna.

## Checklist

Fixes #149


cmejet (approved):
Worked great

PR #197, opened 2021-07-20
ci: add node 16 to matrix

### LINKED ISSUE #150: create-bison-app does not execute on node 16+
Testing the canary release: Node 14 works fine, but Node 16 just installs and exits. We should update our Node matrix to test 16.

kgajera:
`npx create-bison-app@canary myapp` now works on node 16 after https://github.com/echobind/bisonapp/pull/190 was merged. This adds node 16 to the CI matrix.

I wasn't sure if we should also update the template's `.nvmrc` and `.tool-versions` to node 16 or leave it at 14 for now. Let me know if that should be updated too.

## Checklist

Fixes #150


mthomps4:
> I wasn't sure if we should also update the template's .nvmrc and .tool-versions to node 16
If we are confident everything is running with 16 - I'd make the change. 16 is LTS with 18 rounding the corner.

kgajera:
> > I wasn't sure if we should also update the template's .nvmrc and .tool-versions to node 16
> > If we are confident everything is running with 16 - I'd make the change. 16 is LTS with 18 rounding the corner.

I've made this change. No issues noticed running on node 16.

.github/workflows/tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
```
cmejet, line 11: 🎉  YAY, so happy to see this working!

cmejet (approved):
Fantastic to see this working

PR #200, opened 2021-07-21
ci: trigger on pull request event

kgajera:
Currently, our tests workflow is not running on dependabot or public fork PRs so this adds the `pull_request` event as a trigger.

By default, first time contributors should need to get approval before the workflows run:
https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#configuring-required-approval-for-workflows-from-public-forks

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,5 +1,10 @@
 name: Run Tests
-on: [push]
+on:
+  push:
+    branches:
+      - canary
+      - main
+  pull_request:
 
 jobs:
   release:


.github/workflows/tests.yml
```diff
@@ -1,5 +1,10 @@
 name: Run Tests
-on: [push]
+on:
+  push:
+    branches:
```
cmejet, line 4: WIth the previous `on push` we were obviously running each time a commit was pushed, do we have any concern about this? It feels like the right decision given it will run on PR or push to canary/main and prevent using more minutes on git actions.

cmejet (approved):
Looks great! Do we need to do the setup for fist time contributors or was that already done?

.github/workflows/tests.yml
kgajera: I initially had this as `on: [push, pull_request]` but when pushing changes to the PR, it triggered duplicate runs in the CI. So our configuration for this needs to make sure only one event triggers the CI to run at a time. This was the most common configuration I saw others using.

mthomps4 (approved):

PR #210, opened 2021-08-23
chore: upgrade cypress

### LINKED ISSUE #191: Upgrade Cypress
Current version of Cypress as a dev dependency is 2 major versions behind.
We should upgrade to version 8 or newer if at that time it has changed.

image

kgajera:
## Changes

- Upgrades `cypress` and related dependencies
- Fixes path to extended tsconfig file

## Checklist

Fixes #191


packages/create-bison-app/template/tests/e2e/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../cypress/tsconfig.json",
   "compilerOptions": {
     // "baseUrl": "../node_modules",
```
cmejet, line 4: I know this isn't changed, is there a reason this is here & commented out? Should we note why it's there? I know in one of our apps I'm using that line so curious.

cmejet (approved):
Just one question/clarification but not directly related to this change. Feel free to merge.

packages/create-bison-app/template/tests/e2e/tsconfig.json
kgajera: As far as I could tell from the commit history, I don't think that line was ever significant so I've removed it.

PR #221, opened 2021-09-14
chore: upgrade lerna

### LINKED ISSUE #220: chore: Upgrade Lerna
Upgrade Lerna to resolve dependabot alerts. Cannot automatically use dependabot as Lerna depends on older versions of the affected libraries.

kgajera:
Upgrades lerna to the latest major version. Looking at their changelog, I don't think we are affected by any of the breaking changes: https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes

## Checklist

Fixes #220

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,6 @@
   },
   "devDependencies": {
     "conventional-changelog-conventionalcommits": "^4.4.0",
-    "lerna": "^3.22.1"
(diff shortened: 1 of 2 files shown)


mcavaliere (approved):
Looks good to me! :shipit:

PR #224, opened 2021-09-21
chore: upgrade jest

### LINKED ISSUE #207: Upgrade jest
We need to upgrade to the latest major version of jest. They did change some of their default configuration values, specifically `testEnvironment`, which breaks our tests.

https://jestjs.io/blog/2021/05/25/jest-27

kgajera:
## Changes

This upgrades to the latest major version of jest in the template and for the CLI. They have changed the default `testEnvironment` configuration value.

All test files that rely on DOM, now include the following comment specifying the environment to use:
```
/**
 * @jest-environment jsdom
 */
```

From jest docs: https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults
> For mixed projects where some tests require a DOM environment but others don't, we recommend using the fast "node" environment by default and declaring exactly those tests that need the DOM using docblocks.

Next.js's docs suggest doing the same: https://nextjs.org/docs/testing

## Checklist

Fixes #207


mcavaliere (approved):
🎸

PR #230, opened 2021-09-22
ci: run release workflow on node 16

kgajera:
Runs "Release" workflow on node 16 so dependencies can be installed: https://github.com/echobind/bisonapp/runs/3678001603?check_suite_focus=true#step:7:14

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v2.1.2
         with:
-          node-version: 14.6.0
+          node-version: 16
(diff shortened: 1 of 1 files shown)


lschjeide (approved):

.github/workflows/release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v2.1.2
         with:
-          node-version: 14.6.0
+          node-version: 16
```
cmejet, line 21: 🔥 
Thank you!

cmejet (approved):
LGTM

PR #235, opened 2021-10-18
Creates and uses shared Footer component

kyleshevlin:
I noticed an issue with a static year in the footer of the layouts. This PR fixes this, and creates a shared Footer.

## Changes

- Creates a shared `Footer` component
- Dynamically sets the `year` for the copyright
- Changes the text of the footer from `i` to `em` (though arguably this should be done with CSS as I don't think the content itself is semantically emphasized).

## Checklist


cmejet (approved):
Great to see this refactor, one component to rule them all

PR #236, opened 2021-10-19
chore: pin down cypress version to 8.5

kgajera:
Temporary fix for CI cypress failures.

## Checklist


kyleshevlin:
There's a patch version available at `8.3.1`. I don't want to make this more work than necessary, but did you test the patch out?

cmejet (approved):

PR #238, opened 2021-10-20
feat: Make external links themeable with Chakra UI

tundera:
External links are currently implemented with `<a>` tags in JSX. Chakra UI has a `Link` component which has the same functionality as `a`, while also making the links themeable through variants defined in the default or a custom theme via Chakra's theming API. See [this discussion](https://github.com/echobind/bisonapp/pull/235/commits/772a792b8aad287716ccf613fce8244071a75c91#r731948796) for context and an example. 

## Changes

- Replaces use of anchor tags for external links with the `Link` component from Chakra UI
- Replace the use of `target="_blank"` and `rel="noopener noreferrer"` attributes on external links with the `isExternal` prop on the substituted `Link` components which provides the same functionality

## Screenshots

themable-links

## Checklist

### DIFF
--- a/packages/create-bison-app/template/components/Footer.tsx
+++ b/packages/create-bison-app/template/components/Footer.tsx
@@ -1,4 +1,4 @@
-import { Center, Flex, Text } from '@chakra-ui/layout';
+import { Center, Flex, Text, Link } from '@chakra-ui/layout';
 
 import { Logo } from './Logo';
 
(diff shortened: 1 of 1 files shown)


cmejet (commented):
I think there is overlap with this and the footer component PR #235 so we will have some collisions if both are merged.

kyleshevlin (approved):
Looks good, I don't mind if you merge this first, I can fix my PR no problem.

kgajera (commented):
@tundera Can you rebase this with canary? This is good to merge once the CI passes.

PR #243, opened 2021-12-08
feat: add universal chakra link component

tundera:
Description of PR that completes issue here...

## Changes

- Adds a new `Link` component for both dynamic and external links compatible with Next.js and Chakra UI
- Maintains theming support and `as` casting (i.e. `<Button as={Link} href="/" variant="navigation" />`)

## Checklist

### DIFF
--- a/packages/create-bison-app/template/components/Link.tsx
+++ b/packages/create-bison-app/template/components/Link.tsx
@@ -1,0 +1,44 @@
+import type { PropsWithChildren } from 'react';
+
+import NextLink, { LinkProps as NextLinkProps } from 'next/link';
+import { Link as ChakraLink, LinkProps as ChakraLinkProps } from '@chakra-ui/react';
+
+type LinkProps =
+  | PropsWithChildren<NextLinkProps & Omit<ChakraLinkProps, 'as'>>
+  | PropsWithChildren<Omit<NextLinkProps, 'as'> & ChakraLinkProps>;
+
+type NextLinkAs = NextLinkProps['as'];
+type ChakraLinkAs = ChakraLinkProps['as'];
+
+//  Has to be a new component because both chakra and next share the `as` keyword
+export const Link = ({
+  href,
+  as,
+  replace,
+  scroll,
+  shallow,
+  prefetch,
+  locale,
+  isExternal,
+  children,
+  ...chakraProps
+}: LinkProps) => {
+  return isExternal ? (
+    <ChakraLink href={href} as={as as ChakraLinkAs} isExternal {...chakraProps}>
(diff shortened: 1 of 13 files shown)


packages/create-bison-app/template/components/Link.tsx
```diff
@@ -0,0 +1,44 @@
+import type { PropsWithChildren } from 'react';
+
+import NextLink, { LinkProps as NextLinkProps } from 'next/link';
+import { Link as ChakraLink, LinkProps as ChakraLinkProps } from '@chakra-ui/react';
+
+type LinkProps =
```
mcavaliere, line 6: I recommend exporting these types so other components can consume them.

mcavaliere (approved):
Looks good to me, but ask others who are more active on the project to look as well. 

One recommendation: you might want to wrap other clickable components as well for convenience rather than forcing buttons to use the `as` prop - I did something like this for the CITJS website https://github.com/echobind/jamstack-ebook-website/blob/main/components/Link/index.tsx

tundera:
@mcavaliere Cool idea, I like it 💯  I updated the PR to have a convenience wrapper for `Button` and `BreadcrumbLink` as well

packages/create-bison-app/template/layouts/LoggedOut.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Box, Flex, Button } from '@chakra-ui/react';
```
kgajera, line 2: The `Button` import is unused and causing a lint error.

packages/create-bison-app/template/components/Link.tsx
```diff
@@ -0,0 +1,102 @@
+import type { LinkProps, NextLinkAs } from 'types';
```
kgajera, line 1: Is the module for this import right? I'm getting errors when running `yarn dev:typecheck`.

kgajera (changes requested):
Noticed lint and TypeScript errors which are causing the CI failure.

packages/create-bison-app/template/components/Link.tsx
tundera: @kgajera It turned out to be an issue of the types I defined in `types.ts` being in the gitignore file. And even then, the file would have ended up overwritten by GraphQL Codegen generated types (which is why the file is ignored to begin with 😄 ).

The fix was just to move the types I defined for the component in a different file. I ended up declaring them inside the `Link.tsx` component file itself since they weren't being used in any other modules.

kgajera (approved):

PR #244, opened 2022-01-07
feat: Enable TS strict

DominicSherman:
## Changes

- Enable strict mode in the TS config
- Fix all issues that arose from strict being enabled

## Checklist

### DIFF
--- a/packages/create-bison-app/template/components/SignupForm.tsx
+++ b/packages/create-bison-app/template/components/SignupForm.tsx
@@ -66,7 +66,7 @@ export function SignupForm() {
       const { data } = await signup({ variables });
 
       if (!data?.signup?.token) {
-        throw new Error('Signpup failed.');
+        throw new Error('Signup failed.');
       }
 
       await login(data.signup.token);
--- a/packages/create-bison-app/template/graphql/modules/profile.ts
+++ b/packages/create-bison-app/template/graphql/modules/profile.ts
@@ -1,5 +1,6 @@
-import { objectType, inputObjectType } from 'nexus';
-import { ApolloError } from 'apollo-server-errors';
+import { inputObjectType, objectType } from 'nexus';
+
+import { NotFoundError } from '../errors';
 
 // Profile Type
 export const Profile = objectType({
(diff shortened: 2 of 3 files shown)


packages/create-bison-app/template/components/LoginForm.tsx
```diff
@@ -51,9 +51,14 @@ export function LoginForm() {
     try {
       setIsLoading(true);
       const { data } = await login({ variables: formData });
+
+      if (!data?.login?.token) {
+        throw new Error('Login failed.');
```
kgajera, line 56: It's probably out of the scope of this PR, but we should create another issue to display some kind of user feedback when this is error is thrown.

packages/create-bison-app/template/components/SignupForm.tsx
```diff
@@ -65,10 +65,14 @@ export function SignupForm() {
 
       const { data } = await signup({ variables });
 
+      if (!data?.signup?.token) {
+        throw new Error('Signpup failed.');
```
kgajera, line 69: Typo: "Signpup" should be "Signup"

packages/create-bison-app/template/graphql/modules/profile.ts
```diff
@@ -12,12 +13,18 @@ export const Profile = objectType({
...
-        return context.prisma.profile
+      resolve: async (parent, _, context) => {
+        const user = await context.prisma.profile
           .findUnique({
             where: { id: parent.id },
           })
           .user();
+
+        if (!user) {
+          throw new ApolloError('User not found', '404');
```
kgajera, line 24: I would recommend creating a [custom error](https://www.apollographql.com/docs/apollo-server/data/errors/#custom-errors), like `NotFoundError`, for this because we will need to use this frequently in real apps and we don't want to have to repeat the error code each time.

Just my opinion, but I would also favor using `NOT_FOUND` as the error code because Apollo doesn't use HTTP response codes for their [built-in errors](https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes).

kgajera (changes requested):
Good work! Found one typo that should be fixed. The other comments were more opinionated so feel free to leave comments if you disagree.

kgajera (approved):

PR #246, opened 2022-02-09
fix: The DATABASE_URL set in .env.test doesn't include a colon between the username and password.

cullylarson:
If you provide a database password during setup, the `DATABASE_URL` set in `.env.test` will not have a colon between the username and password. If no password is provided, it works fine.

## Changes

- [x] Update `_.env.test.ejs` to put a colon between the username and password if a password is provided.

## Checklist


packages/create-bison-app/template/_.env.test.ejs
```diff
@@ -1,3 +1,3 @@
 # ENV vars here override .env when running tests
-DATABASE_URL="postgresql://<%= db.dev.user %><%= db.dev.password %>@<%= db.dev.host %>:<%= db.dev.port %>/<%= db.test.name %>?schema=public"
-APP_SECRET=foo
\ No newline at end of file
+DATABASE_URL="postgresql://<%= db.dev.user %><% if (db.dev.password) { %>:<%= db.dev.password %><% } %>@<%= db.dev.host %>:<%= db.dev.port %>/<%= db.test.name %>?schema=public"
```
cullylarson, line 2: I just copied this from `_.env.ejs`.

kgajera (approved):

PR #247, opened 2022-02-09
feat: Add import path alias.

cullylarson:
Adds a `@/` path alias.

## Changes

- [x] Add path alias to tsconfig.
- [x] Update jest config to include the path alias from tsconfig.
- [x] Update all import paths to work with the new alias.
- [x] Update the `ts-node` npm script to work with the path alias.

## Checklist

- [x] Generating a new app works.
- [x] Jest and e2e tests pass.
- [x] Prisma, nexus, etc. builds run.
- [x] Database migrations, seeds, etc. run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ We're always improving on this, and we welcome suggestions from the community!
 - Customizable [Hygen Templates](https://www.hygen.io/) to generate new files
 - Fully wired up login/signup pages with client and server-side validation.
 - Eslint pre-configured with [Echobind best practices](https://github.com/echobind/eslint-plugin-echobind)
+- Import path alias to the root project folder (`@/`) to avoid the need for long relative import paths.
 
 ## Conventions
 
--- a/packages/create-bison-app/template/graphql/modules/user.ts.ejs
+++ b/packages/create-bison-app/template/graphql/modules/user.ts.ejs
@@ -257,8 +257,9 @@ export const UserCreateInput = inputObjectType({
     t.field('roles', {
       type: list('Role'),
     });
+
     t.field('profile', {
-      type: 'ProfileRelationalCreateInput'
+      type: 'ProfileRelationalCreateInput',
     });
   },
 });
--- a/packages/create-bison-app/template/graphql/modules/user.ts.ejs
+++ b/packages/create-bison-app/template/graphql/modules/user.ts.ejs
@@ -14,7 +14,7 @@ import { UserInputError } from 'apollo-server-micro';
 
 import { hashPassword, appJwtForUser, comparePasswords } from '@/services/auth';
 import { canAccess, isAdmin } from '@/services/permissions';
-import { prismaArgObject } from './helpers';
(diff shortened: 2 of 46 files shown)


packages/create-bison-app/template/graphql/modules/user.ts.ejs
```diff
@@ -257,8 +257,9 @@ export const UserCreateInput = inputObjectType({
     t.field('roles', {
       type: list('Role'),
     });
+
     t.field('profile', {
-      type: 'ProfileRelationalCreateInput'
-    })
+      type: 'ProfileRelationalCreateInput',
+    });
```
cullylarson, line 263: These are changes that ended up in the app after the initial commit (potentially from something like an `eslint --fix`). So just made them here so there aren't uncommitted changes just after creating the app.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -41,7 +41,7 @@
     "test:e2e": "cross-env DOTENV_CONFIG_PATH=.env.test ts-node -r dotenv/config $(yarn bin)/start-server-and-test 'yarn test:server' http://localhost:3001 'yarn cypress:run'",
     "test:e2e:local": "cross-env CYPRESS_LOCAL=true CYPRESS_BASE_URL=http://localhost:3000 cypress open",
     "test:server": "next start --port 3001",
-    "ts-node": "ts-node-dev --project tsconfig.cjs.json",
+    "ts-node": "ts-node-dev --project tsconfig.cjs.json -r tsconfig-paths/register",
```
cullylarson, line 44: This is what makes `ts-node` work with the path alias. Without it, none of the scripts run with `ts-node` would be able to use path aliases.

packages/create-bison-app/template/tsconfig.json
```diff
@@ -17,7 +17,11 @@
     "sourceMap": true,
     "plugins": [{ "name": "nexus/typescript-language-service" }],
     "typeRoots": ["node_modules/@types", "types"],
-    "rootDir": "."
+    "rootDir": ".",
+    "baseUrl": ".",
```
cullylarson, line 21: For some reason `tsconfig-paths/register` won't work without a `baseUrl` set.

kgajera (approved):

cmejet (approved):
Thanks for taking the lead on this one! Only thing I think we should do is provide a blurb in the readme for bison users to understand our move to using alias'd paths.

README.md
```diff
@@ -24,6 +24,7 @@ We're always improving on this, and we welcome suggestions from the community!
 - Customizable [Hygen Templates](https://www.hygen.io/) to generate new files
 - Fully wired up login/signup pages with client and server-side validation.
 - Eslint pre-configured with [Echobind best practices](https://github.com/echobind/eslint-plugin-echobind)
+- Import path alias to the root project folder (`@/`) to avoid the need for long relative import paths.
```
cmejet, line 27: Perfecto! Thanks for adding this.

PR #252, opened 2022-03-25
refactor: simplify logic to parse default prompt answers

kgajera:
This doesn't change any functionality, but starts to refactor the logic in our CLI handler to be simpler.

## Changes

- Refactors to use `lodash` to parse default prompt answers
- Adds tests for CLI

## Checklist


cmejet (approved):
Nice updates, especially the cli tests!

PR #256, opened 2022-04-08
ci: update to not auto fix lint errors

kgajera:
## Changes

- Updates `yarn lint` to not auto fix lint errors. This was being run in the CI, so the lint step would pass even when lint errors exist.
- Adds new `yarn lint:fix` script that can auto fix lint errors
- Updates the generated GraphQL typings file to not be linted anymore because this was added to the gitignore a while ago so there's no point in linting it.
- Upgrades graphql codegen dependencies

## Checklist


packages/create-bison-app/template/package.json.ejs
```diff
@@ -32,7 +32,8 @@
     "g:test:factory": "hygen test factory --name",
     "g:test:request": "hygen test request --name",
     "g:test:util": "hygen test util --name",
-    "lint": "yarn eslint . --ext .ts,.tsx --fix --ignore-pattern tmp",
+    "lint": "yarn eslint . --ext .ts,.tsx --ignore-pattern tmp",
+    "lint:fix": "yarn eslint . --ext .ts,.tsx --fix --ignore-pattern tmp",
```
cmejet, line 36: I prefer this separation! Great add to not just pull out the ability to fix, but to create another distinct script.

cmejet (approved):
Thanks for continuing to suggest improvements.

packages/create-bison-app/template/package.json.ejs
cullylarson: If you want to make sure both scripts do the same thing, except for fixing lint errors, you could do:

```
lint:fix": "yarn lint --fix"
```

cullylarson (approved):
Great idea, Kishan! I like it.

PR #259, opened 2022-05-05
chore: upgrade dependencies

### LINKED ISSUE #237: Cypress 8.6 failing in CI
Relevant error message: https://github.com/echobind/bisonapp/runs/3931865850?check_suite_focus=true#step:14:44

8.6 does contain an upgrade to Electron which is likely the cause: https://docs.cypress.io/guides/references/changelog#8-6-0

kgajera:
## Changes

- Adds 18 to node matrix and removes 12 (at EOL: https://nodejs.org/en/about/releases/)
- Upgrades many dependencies

Fixes #237


.github/workflows/release.yml
```diff
@@ -10,13 +10,13 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
cullylarson, line 13: Pinning latest can create unintended behavior when a new "latest" is released. Not necessarily a deal-breaker for using it, but something to consider.

cullylarson: Sorry, just noticed this was a draft 😅

kgajera: No problem! Appreciate the comment. Pinned it down to 20.04. PR is now ready for review.

cullylarson (approved):
If this runs in Github Actions, it looks good to me!

PR #260, opened 2022-05-13
adding a Comparison scalar and surfacing it in definition block for S…

cBoss1029:
Adding updates that allow for comparison operators `gte`, `gt`, `lte`, `lt` to accept either `string`, `number`, or `object` without throwing TS errors

## Changes

- Added a `Comparison` custom scalar to be used with comparison operators `gte`, `gt`, `lte`, `lt`
- surfaced `compare` method in definition block using `asNexusMethod`

## Checklist

- [x ] Generating a new app works

Fixes #xxx

### DIFF
--- a/packages/create-bison-app/template/graphql/modules/scalars.ts
+++ b/packages/create-bison-app/template/graphql/modules/scalars.ts
@@ -6,9 +6,59 @@ import {
   URLResolver,
 } from 'graphql-scalars';
 import { asNexusMethod } from 'nexus';
+import { GraphQLScalarType, Kind } from 'graphql';
+
+export const ComparisonOperatorScalarType = new GraphQLScalarType({
+  name: 'ComparisonOperator',
+  description: 'Allows values typically used in comparison operators: string, number, Date object',
+  // parseValue and serialize are used to ensure validity
+  parseValue: (value) => {
+    return typeof value === 'object' || typeof value === 'string' || typeof value === 'number' ? value
+      : null
+  },
+  serialize: (value) => {
+    return typeof value === 'object' || typeof value === 'string' || typeof value === 'number' ? value
+      : null
+  },
+  parseLiteral: (ast) => {
+    switch (ast.kind) {
+      case Kind.STRING:
+      case Kind.INT:
+        return ast.value;
+      case Kind.OBJECT: return parseObject(ast);
+      default: return null
+    }
+  }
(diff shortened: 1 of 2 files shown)


kgajera (commented):
Because we are using the new `compare` type in the `StringFilter`, I'm wondering if this change makes sense. What are the use cases where I would want to pass a number, date, etc. when using a filter for "strings":
```ts
export const StringFilter = inputObjectType({
  name: 'StringFilter',
  description: 'A way to filter string fields. Meant to pass to prisma where clause',
  definition(t) {
    t.string('contains');
    t.string('endsWith');
    t.string('equals');
    t.compare('gt');
    t.compare('gte');
    t.list.nonNull.string('in');
    t.compare('lt');
    t.compare('lte');
    t.list.nonNull.string('notIn');
    t.string('startsWith');
  },
});
```

Does it make more sense for us to define a filter for each of our types like below so we don't sacrifice type safety:

```ts
export const StringFilter = inputObjectType({
  name: 'StringFilter',
  description: 'A way to filter string fields. Meant to pass to prisma where clause',
  definition(t) {
    t.string('contains');
    t.string('endsWith');
    t.string('equals');
    t.string('gt');
    t.string('gte');
    t.list.nonNull.string('in');
    t.string('lt');
    t.string('lte');
    t.list.nonNull.string('notIn');
    t.string('startsWith');
  },
});

export const NumberFilter = inputObjectType({
  name: 'NumberFilter',
  description: 'A way to filter number fields. Meant to pass to prisma where clause',
  definition(t) {
    t.int('equals');
    t.int('gt');
    t.int('gte');
    t.list.nonNull.int('in');
    t.int('lt');
    t.int('lte');
    t.list.nonNull.int('notIn');
  },
});

export const DateFilter = inputObjectType({
  name: 'DateFilter',
  description: 'A way to filter date fields. Meant to pass to prisma where clause',
  definition(t) {
    t.date('equals');
    t.date('gt');
    t.date('gte');
    t.list.nonNull.date('in');
    t.date('lt');
    t.date('lte');
    t.list.nonNull.date('notIn');
  },
});
```

mthomps4:
Curious thoughts on the comment above as well.

cBoss1029:
True. I think that approach makes more sense. I'll refactor.

kgajera (approved):

PR #262, opened 2022-05-23
test: add support for svg

### LINKED ISSUE #240: Jest Tests will break if component imports an image
Recently added an image import to the `Logo` component and broke the test for the component. Why?

When Jest attempts to import the file, it does not know what to do with the binary file. These files are unnecessary for tests and should be mocked. I fixed it by adding the following to the config in `jest.config.js`:

```javascript
moduleNameMapper: {
  // Maps the import of various images to a mock file path
  '^.+\\.(jpg|jpeg|png|gif|webp|svg)$': '<rootDir>/__mocks__/fileMock.js',
}
```

You then must also create a `/__mocks__/fileMock.js` file with the following:

```javascript
module.exports = '/test-file-stub';
```

### Useful Resources

- https://stackoverflow.com/questions/46898638/importing-images-breaks-jest-test
- https://nextjs.org/docs/testing#manual-setup-1

SamanthaShackelford:
## Changes

- Added moduleMapper in the jest config to support images
- Added a mock file for svgs

## Checklist

Fixes #240

### DIFF
--- a/packages/create-bison-app/tasks/copyFiles.js
+++ b/packages/create-bison-app/tasks/copyFiles.js
@@ -74,6 +74,7 @@ async function copyFiles({ variables, targetFolder }) {
 
     cpy(
       [
+        "__mocks__",
         "_templates",
         ".vscode",
         "chakra",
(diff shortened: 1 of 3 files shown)


PaulGoodDay (approved):
Looks good

mthomps4 (approved):

PR #263, opened 2022-06-08
docs: move readme to create-bison-app package and add symlink to it from root

kgajera:
## Changes

- This moves the readme to `packages/create-bison-app/README.md` and adds a symlink to it from the root of the repository.
   -  The reason for doing so is to have our README displayed on NPM next time we publish: https://www.npmjs.com/package/create-bison-app  

https://github.com/echobind/bisonapp/tree/docs/move-readme-to-create-bison-app-package


cmejet (approved):
Nice!

darleendenno (approved):
TIL ✨

PR #268, opened 2022-07-07
chore: upgrade to cypress 10

kgajera:
## Changes

- Upgrades to Cypress 10 which contained a few breaking changes: https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-version-10-0
   -  The configuration files (`cypress.config.ts`,`cypress/support/`, `cypress/plugins/` were automatically migrated by cypress during the upgrade process.
-  Converts our cypress tests from `.js` files to `.ts`
-  Fixes flaky tests for login page due to running assertions before the API request finishes
-  Even before these changes, there were quite a few TypeScript issues with our Cypress set up due to misconfiguration in our tsconfig files. I think this might have been the reason we were using `js` files. This should resolve those.

## Checklist


packages/create-bison-app/template/cypress/plugins/index.ts
```diff
@@ -68,11 +41,11 @@ export default (on, _config) => {
     disconnectDB: () => {
       return disconnect();
     },
-    factory: ({ name, attrs }) => {
-      const Factory = Factories[`${name}Factory`];
+    factory: ({ name, attrs }: { name: string, attrs: any }) => {
+      const Factory = (Factories as any)[`${name}Factory`];
```
alexanderson1993: How intense do you want the type annotations to be here? If you really wanted to, you could get rid of the `any` by using this as the type for the `name` parameter:

```ts
type FactoryNames = keyof typeof Factories extends `${infer T}Factory` ? T : never;
```

Of course, I doubt it would add that much value - it doesn't add autocompletion to the places where this function is called. So feel free to skip it.

packages/create-bison-app/template/cypress/plugins/index.ts
```diff
@@ -68,11 +41,11 @@ export default (on, _config) => {
...
       return Factory.create(attrs);
     },
-    login: (attrs: Pick<User, 'email' | 'password'>): Promise<LoginTaskObject> => {
+    login: async (attrs: Pick<User, 'email' | 'password'>): Promise<LoginTaskObject> => {
```
alexanderson1993, line 51: Is there a specific reason for changing this to `async`? I suppose it's helpful for making it clear the function returns a promise, but the return type also does that. 

Also, if you wanted to, since this is async the promise chain at the end could be replaced with an `await`. Absolutely not a necessary change, of course.

packages/create-bison-app/template/cypress/support/commands.ts
```diff
@@ -6,29 +13,14 @@ import { LOGIN_TOKEN_KEY } from '@/constants';
 import { LoginTaskObject } from '@/cypress/plugins';
 
 declare global {
-  // eslint-disable-next-line
```
alexanderson1993, line 9: 👏

packages/create-bison-app/template/tests/e2e/login.cy.ts
```diff
@@ -1,15 +1,20 @@
+import { User } from '@prisma/client';
+
 describe('Login', () => {
   describe('with an email that doesnt exist', () => {
     it('shows an error', () => {
       const email = 'nowayshouldIexist@wee.net';
       const password = 'test1234';
 
+      cy.intercept('POST', '/api/graphql').as('loginMutation');
```
alexanderson1993, line 9: This is a slick fix. I saw your TIL in Discord, and learned something myself!

alexanderson1993 (approved):
Solid improvements. A few comments, but I'm new here and have no context about anything so feel free to ignore them. 😅

packages/create-bison-app/template/cypress/plugins/index.ts
kgajera: Great suggestion! Updated.

packages/create-bison-app/template/cypress/plugins/index.ts
kgajera: `async` was suggested by VS Code :sweat_smile: The promise chain has been refactored to use `await`.

PR #270, opened 2022-07-08
chore: upgrade to prisma 4

kgajera:
## Changes

- Upgrades to prisma 4
- Sets default value for User `roles` field now that prisma supports setting default value for scalar list fields: https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4#scalar-list-defaults

## Checklist


alexanderson1993 (approved):
Looks great!

PR #273, opened 2022-08-19
chore: Replace the yarnWithEnv script with dotenv-cli.

cullylarson:
There are a few reasons for this change:

1. We no longer need to maintain the `yarnWithEnv` script. This probably wouldn't have been a big deal since it's unlikely we'd need to change it.
2. The `yarnWithEnv` script is potentially a strange thing to have in someone's repo. If someone creates a Bison app, it might not make sense why `yarnWithEnv` is there or how it's working.
3. The current setup, with setting `DOTENV_CONFIG_PATH=.env.test`, makes it impossible to override values in `.env.test` with a `.env.test.local` file (e.g. if you want to use a different database url, etc.). 

## Changes

- Remove the `yarnWithEnv` script and replace it with `dotenv-cli`

## Checklist


kgajera (approved):
👍🏽 Nice improvement!

PR #279, opened 2022-09-23
fix: Login session lifetime could be shorter than expected.

cullylarson:
In `context/auth`, since `now` is defined outside of the component, `COOKIE_EXPIRE_DATE` will always be the same value as long as the user doesn't refresh the browser. So if we allow users to stay logged in for 30 days, a user loads the page, keeps their browser open for 29 days, then logs in, their session will only last 1 day.

Probably not a huge deal (especially since our default session lifetime is 1 year). But it could be an issue if someone wants to use very short sessions.

## Changes

- This PR computes the session lifetime on login, when the token cookie is set.

I generated an app and logged in. The session lifetime is still one year, as expected.

## Checklist

### DIFF
--- a/packages/create-bison-app/template/context/auth.tsx
+++ b/packages/create-bison-app/template/context/auth.tsx
@@ -6,9 +6,9 @@ import { useMeLazyQuery, User } from '@/types';
 import { FullPageSpinner } from '@/components/FullPageSpinner';
 import { LOGIN_TOKEN_KEY } from '@/constants';
 
-const now = new Date();
-const timeValidInMs = 365 * 24 * 60 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


packages/create-bison-app/template/context/auth.tsx
```diff
@@ -63,7 +66,7 @@ function AuthProvider({ ...props }: Props) {
    * Logs out a user by removing their token from cookies.
    */
   async function logout() {
-    cookies().remove(LOGIN_TOKEN_KEY, { path: '/', expires: COOKIE_EXPIRE_DATE });
+    cookies().remove(LOGIN_TOKEN_KEY, { path: '/' });
```
cullylarson, line 69: We don't need to set `expires` when removing the cookie.

cball (approved):

PR #280, opened 2022-09-23
chore: Remove unnecessary React imports.

cullylarson:
Since we're using Next.js we don't need to import `React` in every `.tsx` file. It's done automatically.

## Changes

- Remove all unnecessary `import React from "react";`
- Add `tsconfig.tsbuildinfo` to .gitignore. I've been noticing this file pop up after generating types.

## Checklist

### DIFF
--- a/packages/create-bison-app/template/_.gitignore
+++ b/packages/create-bison-app/template/_.gitignore
@@ -15,6 +15,9 @@ out/
 # nexus
 .nexus
 
+# typescript
+tsconfig.tsbuildinfo
+
 # production
 /build
 
--- a/packages/create-bison-app/template/jest.config.js
+++ b/packages/create-bison-app/template/jest.config.js
@@ -27,7 +27,7 @@ module.exports = {
   globals: {
     'ts-jest': {
       tsconfig: {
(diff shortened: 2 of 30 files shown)


packages/create-bison-app/template/_.gitignore
```diff
@@ -15,6 +15,9 @@ out/
 # nexus
 .nexus
 
+# typescript
+tsconfig.tsbuildinfo
```
cullylarson, line 19: I've noticed this file being generated. Always have to ignore it.

packages/create-bison-app/template/jest.config.js
```diff
@@ -27,14 +27,14 @@ module.exports = {
   globals: {
     'ts-jest': {
       tsconfig: {
-        jsx: 'react',
+        jsx: 'react-jsx',
```
cullylarson, line 30: This fixes an error that was thrown after removing all the explicit `React` imports:

> ‘React’ refers to a UMD global, but the current file is a module. Consider adding an import instead

cball (approved):

PR #282, opened 2022-10-20
feat: Migrate the backend from GraphQL to tRPC

alexanderson1993:
## Changes

This is a major architectural change that's difficult to summarize. I decided to do the entire migration in a single PR since GraphQL itself is so deeply integrated with the rest of Bison. A good portion of the changes are documentation, though, so hopefully reviewing everything else isn't too bad.

First, this PR removes code, config, and dependencies related to GraphQL. That includes Nexus, GraphQL Codegen, Apollo Server, and Apollo Client. This also includes anything that references the files generated by Nexus and GQL Codegen.

Next, I created a tRPC router and middleware for authenticated procedures. I created the `userRouter` and replicated all of the queries and mutations that existed in the GraphQL schema. After wiring it up to an API route, I set up the tRPC client and replaced all the GraphQL queries with tRPC calls, making adjustments as necessary to make everything behave nicely. 

I also rewrote the test files using a new `trpcRequest` helper, which lets you optionally pass a user record to make the request.

Once I confirmed the app was generating correctly, I updated the hygiene templates to use tRPC, updated the ReadMe and other documentation.

## Checklist


packages/create-bison-app/template/context/auth.tsx
```diff
@@ -17,63 +17,50 @@ const AuthContext = createContext<AuthContextObject>({
...
-  // Load current user if there's a token
   useEffect(() => {
-    if (called) return;
+    // We set the token in useEffect to make sure the page properly
+    // hydrates, since the server doesn't have access to the cookie
+    // at this point in the rendering process.
+    setToken(cookies().get(LOGIN_TOKEN_KEY));
+  }, []);
+
+  const meQuery = trpc.user.me.useQuery(undefined, { enabled: !!token });
```
alexanderson1993, line 30: `{ enabled: !!token }` only fetches the user record if there is a valid token.

packages/create-bison-app/template/jest.config.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
       },
     },
   },
+  transformIgnorePatterns: ['/node_modules/(?!(@swc|@trpc))/'],
```
alexanderson1993, line 34: This is necessary for tRPC to work properly. Normally, Jest doesn't transform node_modules, but one of the tRPC packages has some syntax that Jest doesn't understand natively. This makes it so only the tRPC stuff is transformed by Jest.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -21,21 +18,20 @@
     "db:seed": "yarn prisma db seed",
     "db:seed:prod": "cross-env APP_ENV=production prisma db seed",
     "db:setup": "yarn db:reset",
-    "dev": "concurrently -n \"WATCHERS,NEXT\" -c \"black.bgYellow.dim,black.bgCyan.dim\" \"yarn watch:all\" \"next dev\"",
+    "dev": "next dev",
```
alexanderson1993, line 21: We don't need to run TypeScript concurrently with our dev server; IDEs take care of TypeScript for us.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -44,42 +40,33 @@
...
   },
   "dependencies": {
-    "@apollo/client": "^3.6.1",
     "@chakra-ui/react": "^2.2.0",
     "@chakra-ui/theme": "^2.1.0",
     "@emotion/react": "^11.0.0",
     "@emotion/styled": "^11.0.0",
     "@prisma/client": "^4.0.0",
-    "apollo-server-micro": "^3.6.7",
+    "@tanstack/react-query": "^4.8.0",
```
alexanderson1993: We probably should bump this to the latest version, or even wait to merge this until tRPC 10 is out of beta. Though in my testing it's been very stable.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -88,14 +75,13 @@
...
     "@types/jest": "^26.0.6",
+    "@types/jsonwebtoken": "^8.5.9",
     "@types/node": "^16.9.6",
     "@types/pg": "^8.6.1",
     "@types/react": "^18.0.14",
-    "@types/supertest": "^2.0.10",
     "@vercel/node": "^1.7.3",
     "chance": "^1.1.8",
     "commander": "^8.1.0",
-    "concurrently": "^7.1.0",
```
alexanderson1993, line 98: Not used anymore

packages/create-bison-app/template/package.json.ejs
```diff
@@ -110,13 +96,11 @@
     "jest": "^27.2.1",
     "jest-environment-node": "^27.2.0",
     "lint-staged": "^10.2.11",
-    "msw": "^0.20.1",
```
alexanderson1993, line 113: Not used anywhere in the codebase

packages/create-bison-app/template/package.json.ejs
```diff
@@ -110,13 +96,11 @@
...
     "nanoid": "^3.1.10",
     "pg": "^8.3.0",
     "prettier": "^2.6.2",
     "prisma": "^4.0.0",
     "start-server-and-test": "^1.14.0",
-    "supertest": "^4.0.2",
```
alexanderson1993, line 119: Not used anymore

packages/create-bison-app/template/server/trpc.ts
```diff
@@ -0,0 +1,35 @@
...
+      data: {
+        ...shape.data,
+        invalidArgs: error instanceof BisonError ? error.invalidArgs : undefined,
+        message: error.message,
+      },
+    };
+  },
+});
+
+export class BisonError extends TRPCError {
```
alexanderson1993, line 22: This special error class makes it possible for errors sent to the client to include metadata. The `invalidArgs` comes from the GraphQL extension we were using for this same behavior. Other fields could be added if we wanted to extend it further.

packages/create-bison-app/template/tests/utils.tsx
```diff
@@ -17,58 +23,29 @@ export { default as userEvent } from '@testing-library/user-event';
  */
 export function render(ui: RenderUI, { router = {}, ...options }: RenderOptions = {}) {
   return defaultRender(ui, {
-    wrapper: ({ children }) => {
+    wrapper: function Wrapper({ children }) {
+      const [queryClient] = useState(() => new QueryClient());
+
+      const [trpcClient] = useState(() =>
+        trpc.createClient({
```
alexanderson1993, line 30: One thing this doesn't do is provide a way to mock requests. We could add that feature here - a special link which takes inputs and returns some static output provided in the body of the test.

docs/trpc-examples.md
```diff
@@ -0,0 +1,831 @@
...
+          .findUnique({
+            where: { id: parent.id },
+          })
+          .profile();
+      },
+    });
+
+    t.list.field('posts', {
+      type: 'Post',
+      resolve: async (parent, _, context) => {
```
cball: This is all nexus correct?

docs/trpc-examples.md
```diff
@@ -0,0 +1,831 @@
...
+export const AuthPayload = objectType({
+  name: 'AuthPayload',
+  description: 'Payload returned if login is successful',
+  definition(t) {
+    t.field('user', { type: 'User', description: 'The logged in user' });
+    t.string('token', {
+      description: 'The current JWT token. Use in Authentication header',
+    });
+  },
+});
```
cball: also nexus

docs/trpc-examples.md
```diff
@@ -0,0 +1,831 @@
...
+
+export const userRouter = t.router({
+  me: protectedProcedure.query(async ({ ctx }) => ctx.user),
+  users: protectedProcedure
```
cball: I know this is just documentation, but as a general convention, do you think it's worth putting a comment above each procedure explaining what it is?

Example:

```ts
/*
   Returns a list of users with profiles. Ordered name desc by default.
*/
```

Most of these are straightforward by reading through or with really descriptive procedure names, but it could help with scanning the file and when autocompleting with TS?

docs/trpc-examples.md
```diff
@@ -0,0 +1,831 @@
...
+    .query(async ({ ctx, input }) => {
+      const { where } = input;
+
+      return await ctx.db.user.findUnique({
+        where,
+        select: defaultUserSelect,
+        include: { profile: true, posts: true },
+      });
+    }),
+});
```
cball, line 353: 👍 great points

docs/trpc-examples.md
```diff
@@ -0,0 +1,831 @@
...
+
+      return await ctx.db.user.findMany({
+        where: updatedQuery,
+        orderBy,
+        select: defaultUserSelect,
+        include: { profile: true },
+      });
+    }),
+    // ...
+})
```
cball, line 420: 👍

packages/create-bison-app/README.md
```diff
@@ -29,8 +28,7 @@ We're always improving on this, and we welcome suggestions from the community!
 ## Conventions
 
 - Don't copy/paste files, use generators and Hygen templates. Need to update the generator as your project evolves? they are all in the `_templates` folder.
-- Use a single command to run Next, generate Nexus types, and GraphQL types for the frontend. By doing this you can ensure your types are always up-to-date.
-- Don't manually write types for GraphQL responses... use the generated query hooks from GraphQL Codegen.
+- Don't manually write types for tRPC procedures. Infer the types from the router definition.
```
cball, line 31: 👍

packages/create-bison-app/scripts/createDevAppAndStartServer.js
```diff
@@ -97,20 +97,12 @@ async function createTemplateSymlinks(appPath) {
...
-  const unlinkFile = async (filename) => {
-    const filePath = templatePath(filename);
-    const lstat = fs.lstatSync(filePath, { throwIfNoEntry: false });
-    if (lstat && lstat.isSymbolicLink()) {
-      await fs.promises.unlink(filePath);
-    }
-  };
-
-  await unlinkFile("api.graphql");
-  await unlinkFile("types.ts");
```
cball, line 109: 🎉

packages/create-bison-app/template/lib/trpc.ts
```diff
@@ -0,0 +1,62 @@
...
+import { createTRPCNext } from '@trpc/next';
+import { httpBatchLink } from '@trpc/client';
+
+import { cookies } from './cookies';
+
+import { LOGIN_TOKEN_KEY } from '@/constants';
+import type { AppRouter } from '@/server/routers/_app';
+
+// exporting this so that we consistently use the same transformer everywhere.
+export const transformer = superjson;
```
cball, line 12: 👍

cball (changes requested):
Looks great! 🎉 for major deletions.

The only thing I noticed was a bit of old GQL cruft in the readme, but otherwise things look great.

Also just a quick question on procedure comment conventions. If we want to adopt that, it might be good to add them to the hygen template.

docs/trpc-examples.md
alexanderson1993: Yeah, I must have missed this.

docs/trpc-examples.md
alexanderson1993: Yeah, or we could name the procedures better. The entire router is called `user` so the procedures could be called `getAll`, `getById` or `getMe`, and accessed like

```
trpc.user.getAll.useQuery()
trpc.user.getById.useQuery()
trpc.user.getMe.useQuery()
```

We've got options, I'm happy to adjust or add comments.

cball: We can also do this is a followup. just came to mind as I was reviewing. I feel like the right answer is probably both in some capacity.

packages/create-bison-app/template/package.json.ejs
mthomps4: We may want to use something like this to mock NextAuth later. 
But agree -- remove for now and we can determine what's needed.

packages/create-bison-app/template/tests/utils.tsx
mthomps4: 🤔 This may be another place where `msw` could come into play. 
I'll look at the testing pieces with @dennis-campos when reviewing NextAuth

mthomps4 (approved):

packages/create-bison-app/template/tests/utils.tsx
cball: I think that's a good thing, as it discourages mocking until we absolutely need it.

alexanderson1993: What would be really cool is if we had bulletproof Stripe/Twilio mocks and SQLite support in Bison. Then, tests run e2e without needing a test database running, and only our third-party APIs are mocked.

alexanderson1993: We could set up tRPC to use a special client-side link (instead of the BatchHTTPLink) that calls the tRPC procedures directly instead of over HTTP

kgajera (approved):

prestonkelly1 (approved):

PR #283, opened 2022-10-26
Trpc nextauth

dennis-campos:
Adds NextAuth to work with tRPC.

This PR includes the credentials provider.


packages/create-bison-app/template/constants.ts
```diff
@@ -3,3 +3,5 @@ export const EMAIL_REGEX = /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i;
 
 // The key used to store the login token in the auth cookie
 export const LOGIN_TOKEN_KEY = 'myapp-user';
```
mthomps4, line 5: 1. are we using this anywhere? I don't think so...
2. If we are -- Seeing this is a template file, we should place the real app name here (one less thing to change)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cSpell.words": ["trpc"]
```
alexanderson1993: This seems unnecessary.

packages/create-bison-app/template/cypress/plugins/index.ts
```diff
@@ -48,14 +42,15 @@ export const setupNodeEvents: Cypress.ConfigOptions['setupNodeEvents'] = (on, _c
...
-      const response = await trpcRequest().user.login(variables);
-      const { token } = response;
-      cookies().set(LOGIN_TOKEN_KEY, token);
-      return { token };
+    mockValidSession: ({ user }: { user: NextAuthSession['user'] }): NextAuthSession => {
+      const isAdmin = user.roles.includes(Role.ADMIN);
+      return {
+        user,
+        isAdmin,
+        expires: new Date(Date.now() + 2 * 86400).toISOString(),
```
alexanderson1993: Nit: I always prefer to calculate out my durations to make it clear what the duration is.

Also, the addition appears to be in seconds,  but `Date.now()` returns milliseconds, so this should be altered anyway.

```suggestion
        expires: new Date(Date.now() + 2 * 24 * 60 * 60 * 1000).toISOString(),
```

packages/create-bison-app/template/pages/_app.tsx
```diff
@@ -1,42 +1,22 @@
...
-import type { AppProps } from 'next/app';
-import dynamic from 'next/dynamic';
+import { Session } from 'next-auth';
+import React from 'react';
+import type { AppProps as NextAppProps } from 'next/app';
 
 import { AllProviders } from '@/components/AllProviders';
-import { useAuth } from '@/context/auth';
-import { trpc } from '@/lib/trpc';
+// import '@fontsource/inter/variable.css';
```
alexanderson1993: Might as well remove this entirely

```suggestion
```

packages/create-bison-app/template/pages/index.tsx
```diff
@@ -1,7 +1,14 @@
 import Head from 'next/head';
 import { Heading, Center } from '@chakra-ui/react';
+import { useSession } from 'next-auth/react';
 
 function Home() {
+  const { data: session } = useSession();
+
+  const name = session?.user?.profile?.firstName;
+
+  const welcomeMsg = name ? `Welcome, ${name}!` : 'Welcome, Guest!';
```
alexanderson1993, line 10: 👨‍🍳😘

packages/create-bison-app/template/utils/getDefaultServerSideProps.ts
```diff
@@ -0,0 +1,14 @@
...
+import { unstable_getServerSession } from 'next-auth';
+
+import { authOptions as nextAuthOptions } from '@/pages/api/auth/[...nextauth]';
+
+// Next API route example - /pages/api/restricted.ts
+export const getServerAuthSession = async (ctx: {
+  req: GetServerSidePropsContext['req'];
+  res: GetServerSidePropsContext['res'];
+}) => {
+  return await unstable_getServerSession(ctx.req, ctx.res, nextAuthOptions);
```
alexanderson1993, line 13: 🤩

packages/create-bison-app/template/components/auth/LoginForm.tsx
```diff
@@ -0,0 +1,133 @@
+import {
```
alexanderson1993, line 1: Screenshot 2022-11-02 at 3 50 30 PM

Eventually we'll want to make this UI look better in dark mode, though I'm not considering that necessary for this PR.

alexanderson1993 (changes requested):
Pretty critical bug on the cookie duration. A few other suggestions. Overall, looks good. Merge the suggestion on the cookie, and I'll approve it.

kgajera (changes requested):
I think we're missing a migration for the `Profile.image` column:
Screen Shot 2022-11-02 at 3 55 54 PM

I'm in favor of just deleting all the migrations and creating a new "initial" migration.

alexanderson1993:
I agree with @kgajera - We should probably create a brand new migration for the entire app.

mthomps4:
> I think we're missing a migration for the `Profile.image` column: 
> I'm in favor of just deleting all the migrations and creating a new "initial" migration.

Weird -- That should have been added here: packages/create-bison-app/template/prisma/schema.prisma
Looking back at the migration I copied over -- It doesn't look like it matches. 

I'll work to make a new initial migration. 

I'm assuming the easiest way is to run this fresh from `yarn dev db:migrate` and copy over to the template?

.vscode/settings.json
mthomps4: My B -- I got tired of seeing the Blue Squiggly
Must have hit Workspace instead of User

packages/create-bison-app/template/components/auth/LoginForm.tsx
mthomps4: Yea, I need to get a PR rolling with Chakra Theme files -- but I'm waiting on some other conversations (RFCs) to land first. 

@alexanderson1993 😉

alexanderson1993: I'll be in your DMs

dennis-campos: Yes, I noted this and felt the same about it - I did not want to include it in this PR. I will create PR when this PR gets merged in.

mthomps4:
Realized my error -- I forgot to clean dev 
I updated the schema for the template but not the schema when running migrations.

Incoming 🏃🏼‍♂️

.vscode/settings.json
mthomps4: Removed here: [32e728e](https://github.com/echobind/bisonapp/pull/283/commits/32e728eb0766be7d4506cdea12f23ab1a86fc89a)

alexanderson1993 (approved):

kgajera (approved):

PR #286, opened 2022-11-03
chore: use yarn install --immutable instead of --frozen-lockfile.

cullylarson:
## Changes

- `--frozen-lockfile` is deprecated. It will be removed in a future release. c.f. https://yarnpkg.com/cli/install.
- Rename `main.js.yml.ejs` to `main.yml.ejs` (why does it need `.js` in the filename?)

## Checklist


.github/workflows/tests.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install bison app packages
-        run: yarn install
+        run: yarn install --immutable
```
cullylarson, line 89: Was there a reason we weren't using `--immutable` here? Maybe not needed?

kgajera: The `yarn.lock` for the test bison app is created in the CI so it's not needed but I think this is fine to add here.

kgajera (approved):

mthomps4 (approved):

PR #288, opened 2022-11-09
feat: Markdown standard for client docs

mthomps4:
## Changes
- Adds `.markdownlint.json`
  - Adds default set of rules 
- Adds settings and extension recommend 
- Updates root Readme to adhere to new rules 

## Why 
I don't really care which rules get thrown into the `.markdownlint` file. 
My goal here is to have _something*_ to start with. 
I  personally add this to every client project when creating `.md` docs for handoff.

### DIFF
--- a/packages/create-bison-app/template/README.md.ejs
+++ b/packages/create-bison-app/template/README.md.ejs
@@ -1,7 +1,6 @@
 # <%= name %>
 
 <p align="center" style="text-align:center">
-  <p><%= name %></p>
   <p><img alt="CI STATUS" src="https://github.com/<OWNER>/<REPOSITORY>/actions/workflows/main.yml/badge.svg"/></p>
 </p>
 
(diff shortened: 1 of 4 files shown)


packages/create-bison-app/template/README.md.ejs
```diff
@@ -1,25 +1,37 @@
+# <%= name %>
+
 <p align="center" style="text-align:center">
-  <h1><%= name %></h1>
+  <p><%= name %></p>
```
mthomps4: In a perfect world, we'd throw something like a description in here -- but that's overkill for this PR.

alexanderson1993 (approved):

kgajera (approved):

packages/create-bison-app/template/README.md.ejs
kgajera: I think we could just remove the `<p><%= name %></p>`.

PR #293, opened 2022-12-10
enhancement: Normalize *.local use to avoid confusion -- add NextAuth placeholders

mthomps4:
## Details 
I had a "user issue" with Playwright where it was nabbing a different local env than expected—updating here to stay consistent when rolling new apps. 

Rather than setting bogus `.env.example` files these examples now live with the dotfiles and `*.local` is used as the correct override. 

**PROD:** 
.env .env.local 
**DEV:** 
.env .env.development.local
**TEST:** 
.env.test .env.test.local

## Changes
- Avoid confusion and add consistency with *.local files 
- Add NextAuth placeholders


packages/create-bison-app/template/_.env.ejs
```diff
@@ -1,7 +1,12 @@
+# Copy to .env.local to mimic production
+
 # Environment variables declared in this file are automatically made available to Prisma.
 # See the documentation for more detail: https://pris.ly/d/prisma-schema#using-environment-variables
 
 # Prisma supports the native connection string format for PostgreSQL, MySQL and SQLite.
 # See the documentation for all the connection string options: https://pris.ly/d/connection-strings
 
 DATABASE_URL="postgresql://<%= db.dev.user %><% if (db.dev.password) { %>:<%= db.dev.password %><% } %>@<%= db.dev.host %>:<%= db.dev.port %>/<%= db.dev.name %>?schema=public"
```
mthomps4, line 9: CI Build test still looked at `.env` DATABASE_URL 
Added Dev line to pass.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -31,7 +31,7 @@
     "lint": "yarn eslint . --ext .ts,.tsx --ignore-pattern tmp",
     "lint:fix": "yarn lint --fix",
     "run:script": "yarn ts-node prisma/scripts/run.ts -f",
-    "setup:dev": "yarn db:deploy && yarn db:seed",
+    "setup:dev": "yarn build:prisma && yarn db:deploy && yarn db:seed",
```
mthomps4, line 34: @kgajera - Added these in from your comments last call.

PR #295, opened 2022-12-11
feat: Playwright > Cypress

mthomps4:
image

## Changes
- Takes all the Playwright setup and learnings from a previous project 
- Kills Cypress 
- New CI Flows 

## Checklist


.github/workflows/tests.yml
```diff
@@ -26,6 +26,9 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: Install Yarn
+        run: npm install -g yarn
+
```
mthomps4: Added this when trying to test locally with `act` 
Can remove...

packages/create-bison-app/template/.github/workflows/main.yml.ejs
```diff
@@ -64,31 +64,26 @@ jobs:
...
-        # Only capture images on failure
-        if: failure()
-        with:
-          name: cypress-screenshots
-          path: cypress/screenshots
-      - name: Upload Cypress Videos
-        uses: actions/upload-artifact@v1
-        # Test run video was always captured, so this action uses "always()" condition
+          NEXTAUTH_URL: ${{ secrets.NEXTAUTH_URL }}
+          NEXTAUTH_SECRET: ${{ secrets.NEXTAUTH_SECRET }}
```
mthomps4: 🎗️ We probably need to add docs for this when creating a new app. (Git Secrets)

packages/create-bison-app/template/tests/factories/user.ts
```diff
@@ -53,12 +53,13 @@ export const UserFactory = {
   upsert: async ({ where, createArgs = {}, updateArgs = {}, select = {} }: FactoryUpsertArgs) => {
     // Grab Build Defaults for Create
     const userArgs = UserFactory.build(createArgs, select);
+    const password = updateArgs.password ? hashPassword(updateArgs.password as string) : undefined;
```
mthomps4, line 56: I was upserting the password with raw text on consecutive runs... 🤦🏼

packages/create-bison-app/template/.github/workflows/main.yml.ejs
mthomps4: I'm going to add a Google SSO example (and possibly others) that aren't Email/Password. 
I'll do a whole write-up with that PR next.

kgajera: I'm wondering if these are really needed here? If we are starting the server with `yarn withEnv:test`, shouldn't these variables exist from the test .env file.

packages/create-bison-app/template/.github/workflows/playwright.example
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
kgajera: I think we could delete this file because it's already in `main.yml`

packages/create-bison-app/template/.github/workflows/main.yml.ejs
mthomps4: @kgajera Good call out -- I can nix this for Email/Password 
In the next iteration -- I'll need an actual Google Key, etc. 
Only the placeholders would live in the example dotfiles for that. 

I have both in my spike. 

--- 

I am removing this from CI tho and giving it a go.

packages/create-bison-app/template/tests/e2e/playwrightExamples/example.play.ts
```diff
@@ -0,0 +1,20 @@
+import { test, expect } from '@playwright/test';
```
kgajera, line 1: These examples might be ok to delete unless we think they are helpful beyond the other tests we have.

.github/workflows/tests.yml
kgajera: Agree we should remove if it was only for `act`.

kgajera (approved):
Looks great. Left a few non-blocking comments to remove extra things unless you think they would be helpful to have in the repo.

packages/create-bison-app/template/.github/workflows/playwright.example
mthomps4: :+1:

packages/create-bison-app/template/tests/e2e/playwrightExamples/example.play.ts
mthomps4: I was torn -- the tests we have barely do anything. Simply test that the `storageSession` worked. 
If it's okay, I may leave for now until I get some broader E2E tests with NextAuth SSO in.

PR #308, opened 2023-04-07
Tailwind

alexanderson1993:
## Changes

- Removes all references to Chakra-UI, emotion, and other CSS-in-JS in the template
- Adds Tailwind, along with the necessary components from ui.shadcn.com
- Updates templates and readme accordingly

## Screenshots

Screenshot 2023-04-07 at 1 12 46 PM

## Checklist


packages/create-bison-app/tasks/copyFiles.js
```diff
@@ -39,6 +39,17 @@ async function copyFiles({ variables, targetFolder }) {
...
+    copyWithTemplate(
+      fromPath("tailwind.config.js"),
+      toPath("tailwind.config.js"),
+      variables
+    ),
+    copyWithTemplate(
+      fromPath("postcss.config.js"),
+      toPath("postcss.config.js"),
+      variables
+    ),
```
kgajera: Can these files be moved down to be copied by the `cpy` function: https://github.com/echobind/bisonapp/blob/canary/packages/create-bison-app/tasks/copyFiles.js#L89

The `copyWithTemplate` is only necessary to copy EJS templates.

packages/create-bison-app/template/package.json.ejs
```diff
@@ -46,25 +46,25 @@
...
     "@tanstack/react-query-devtools": "^4.16.1",
     "@trpc/client": "^10.0.0",
     "@trpc/next": "^10.0.0",
     "@trpc/react-query": "^10.0.0",
     "@trpc/server": "^10.0.0",
     "bcryptjs": "^2.4.3",
+    "class-variance-authority": "^0.5.1",
+    "clsx": "^1.2.1",
     "cross-fetch": "3.0.5",
     "framer-motion": "^8.5.0",
```
kgajera: `framer-motion` was also only installed because it was required for chakra. I think we can remove.

packages/create-bison-app/template/tailwind.config.js
```diff
@@ -0,0 +1,33 @@
+/** @type {import('tailwindcss').Config} */
```
kgajera, line 1: I saw Tailwind now supports TypeScript config files: https://tailwindcss.com/blog/tailwindcss-v3-3#esm-and-typescript-support

kgajera (changes requested):
This is great! Left a few comments.

packages/create-bison-app/template/tailwind.config.js
alexanderson1993: I'm gonna stick with .js, since with the JSDoc comment it provides the same benefits as .ts, without the potential for tooling headaches.

kgajera (approved):

PR #311, opened 2023-04-07
feat: Add a health check endpoint.

cullylarson:
Most of the hosting services we use (Render and Flightcontrol) require a health check endpoint. This PR adds such an endpoint that also includes some helpful info.

## Checklist

### DIFF
--- a/packages/create-bison-app/template/pages/api/health.ts
+++ b/packages/create-bison-app/template/pages/api/health.ts
@@ -1,0 +1,30 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+import { prisma } from '@/lib/prisma';
+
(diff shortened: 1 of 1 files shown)


kgajera:
Should we refactor this to use the config changes from https://github.com/echobind/bisonapp/pull/312?

cullylarson:
> Should we refactor this to use the config changes from #312?

@kgajera The only value from `config` that I've used in the health endpoint on other projects is `stage`. Most of the rest of the ones included here aren't directly referenced in `config`. E.g. `NODE_ENV` and `APP_ENV` are consolidated into `stage`. So it's helpful to see the values directly.

kgajera (approved):

PR #312, opened 2023-04-07
feat: Centralize configuration in one place.

cullylarson:
## Changes

- Move all (most) `process.env` references into `config.ts`.
- Add zod checks for config.
- Remove `ts-node-dev`. Doesn't seem to be needed, creates some weird situations where processes don't exist (e.g. `buildProd`).
- Refine the `buildProd` script.
- fix: Prettier config in `bisonapp` is different from prettier config in `packages/create-bison-app/template`. So if you save a file while editing `bisonapp` and then create a bison app from it, it may have lint errors. Resolved this by copying the prettier config from `packages/create-bison-app/template` into the root folder.

## Checklist


packages/create-bison-app/template/config.ts
```diff
@@ -0,0 +1,97 @@
...
+// listed in next.config.js will be available on the client.
+export const config = configSchema.parse({
+  stage,
+  ci: {
+    isCi: process.env.CI,
+    isPullRequest: process.env.IS_PULL_REQUEST,
+  },
+  database: {
+    url: process.env.DATABASE_URL,
+    shouldMigrate: process.env.SHOULD_MIGRATE,
```
kgajera, line 85: Should `shouldMigrate` be used here https://github.com/echobind/bisonapp/blob/cully/centralize-config/packages/create-bison-app/template/scripts/buildProd.ts#L12?

packages/create-bison-app/template/config.ts
```diff
@@ -0,0 +1,97 @@
+import { z } from 'zod';
+
+import { notEmpty } from './lib/type-witchcraft';
+
+const stages = ['production', 'staging', 'development', 'test'] as const;
```
kgajera: This is more of a question about when `staging` should be used and the purpose of when it should be used in the codebase. In your experience, do you set `NODE_ENV=staging` for any deploys that are not production? I've always used `production` for these.

kgajera (changes requested):
It looks like this removes the `APP_ENV` and `BASE_URL` variables, we should also remove them from all the env files in the template.

cullylarson:
> It looks like this removes the `APP_ENV` and `BASE_URL` variables, we should also remove them from all the env files in the template.

@kgajera I removed them in [another PR](https://github.com/echobind/bisonapp/pull/312), was going to resolve it once that one is merged.

cullylarson:
> It looks like this removes the `APP_ENV` and `BASE_URL` variables, we should also remove them from all the env files in the template.

@kgajera I removed them in https://github.com/echobind/bisonapp/pull/313. Was going to resolve that once it's merged.

packages/create-bison-app/template/config.ts
cullylarson: I think importing `config.ts` breaks this script, but let me double-check

packages/create-bison-app/template/config.ts
cullylarson: Good point. We probably won't use staging on most apps.

packages/create-bison-app/template/config.ts
cullylarson: That worked. I also made some other refinements to that script.

packages/create-bison-app/template/config.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const configSchema = z.object({
+  stage: z.enum(stages),
+  ci: z.object({
+    isCi: coerceBoolean,
+    isPullRequest: coerceBoolean,
+  }),
+  database: z.object({
+    url: z.string(),
+    shouldMigrate: coerceBoolean,
```
kgajera, line 67: Do we want to default `shouldMigrate` to `true` or `process.env.NODE_ENV === 'production'`? I think this might trip people up.

kgajera (approved):
Looks great! One more non-blocking comment.

packages/create-bison-app/template/config.ts
cullylarson: It's set explicitly by the `SHOULD_MIGRATE` environment variable. Which, I realized I forgot to add to the .env files.

cullylarson:
CI is failing because of some issues with how `PORT` is set in the .env files. Will resolve in #313.

PR #313, opened 2023-04-07
chore: Simplify env files.

cullylarson:
## Changes

- fix: A few `copyFile` entries mistakenly target the wrong files.
- chore: Remove `.env.local`. We shouldn't ever need `.env.local` because it will be applied at every stage (not usually intended).
- chore: Don't override values in `env.*.local` unless needed.
- chore: Don't set any values in `.env`. Use it as basically a template. The value in `.env` will be applied to every environment, which doesn't usually make sense.

## Checklist


alexanderson1993 (approved):

packages/create-bison-app/template/_.env.test.ejs
```diff
@@ -1,10 +1,8 @@
-# Copy to .env.test.local to override
-# Include NODE_ENV for package script withEnv to work as expected
+# Override values in .env.development.local
```
cullylarson: Fixed this. Will push with any other changes.

packages/create-bison-app/template/_.env.ejs
```diff
@@ -1,19 +1,8 @@
-# Copy to .env.local to mimic production
+APP_ENV=
```
kgajera:
```suggestion
```

packages/create-bison-app/template/_.env.development.ejs
```diff
@@ -1,17 +1,8 @@
...
-# Environment variables declared in this file are automatically made available to Prisma.
-# See the documentation for more detail: https://pris.ly/d/prisma-schema#using-environment-variables
-
-# Prisma supports the native connection string format for PostgreSQL, MySQL and SQLite.
-# See the documentation for all the connection string options: https://pris.ly/d/connection-strings
-
-# Include NODE_ENV for package script withEnv to work as expected
+# Override values in .env.development.local
 
 APP_ENV="development"
```
kgajera:
```suggestion
```

kgajera (approved):
Unless I've missed something, we can remove `APP_ENV` after #312 is merged.

PR #315, opened 2023-04-13
chore: Playwright refinements.

cullylarson:
## Changes

- chore: Reset db before all e2e tests.
- refactor: Use Playwright's `baseURL` setting instead of explicitly providing a full URL on all navigations.
- refactor: Don't use `storageState` to log users in. Instead create the user and directly set session cookies when needed.
- refactor: Stop using zod to verify config. It doesn't really work if you want to share config between client/server. The empty values on the client will always cause the schema validation to fail, so you end up having to coerce values, and in that case, there's no real validation going on. Better to just set a type and make sure the values conform to that type.

## Checklist


packages/create-bison-app/template/playwright.config.ts
```diff
@@ -41,10 +43,11 @@ const config: PlaywrightTestConfig = {
   reporter: 'html',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: baseUrl,
```
cullylarson, line 47: With this set, we don't need to pass a full URL to each page navigation.

packages/create-bison-app/test/tasks/copyFiles.test.js
```diff
@@ -1,36 +1,36 @@
-const fs = require("fs");
-const path = require("path");
+const fs = require('fs');
```
cullylarson, line 1: Almost all of the changes in this file are prettier formatting.

packages/create-bison-app/test/tasks/copyFiles.test.js
```diff
@@ -140,142 +134,134 @@ describe("copyFiles", () => {
...
-        "constants.ts",
-        "global-setup.ts",
-        "global-teardown.ts",
-        "tsconfig.json",
+        '.eslintrc.json',
+        'auth.play.ts',
+        'constants.ts',
+        'global-setup.ts',
+        'global-teardown.ts',
+        'tsconfig.json',
```
cullylarson: Removing two entries from this array is the only real change in this file.

packages/create-bison-app/template/tests/e2e/helpers.ts
```diff
@@ -35,3 +70,54 @@ export async function uploadFile(page: Page, clickableTarget: Locator, filePath:
 
   await fileChooser.setFiles(filePath);
 }
+
+export async function loginAs(page: Page, userInitial: Partial<SessionUser>) {
```
cullylarson, line 73: This function creates a user and logs them in by directly setting the session cookie.

packages/create-bison-app/template/tests/e2e/auth.play.ts
```diff
@@ -1,31 +1,39 @@
...
+import { disconnect } from '@/lib/prisma';
+
+test.beforeEach(async () => resetDB());
+test.afterAll(async () => disconnect());
 
 test.describe(() => {
-  test.use({ storageState: USER.storageState });
+  test('Can login as a User', async ({ page }: { page: Page }) => {
+    const firstName = 'Qwerty';
+    await loginAs(page, { roles: [Role.USER], firstName });
```
cullylarson, line 16: At some point, this test should be updated to log the user in using the UI, instead of `loginAs`. But we should have at least one test that uses `loginAs` at that point, so we have an example.

kgajera (approved):

PR #322, opened 2023-07-14
fix: Minor login improvements

dennis-campos:
## Changes

- Fix an issue where the form justified to the left when resizing to a small viewport.
- Fix an issue where the router was not routing them to `/login` page after signing out.
- Fix an issue if you enter your email and password, and you press "enter," it redirects you to `/reset-password` when it should log you in.

## Checklist

### DIFF
--- a/packages/create-bison-app/template/src/components/auth/LoginForm.tsx
+++ b/packages/create-bison-app/template/src/components/auth/LoginForm.tsx
@@ -1,7 +1,6 @@
 import { useState } from 'react';
 import { signIn } from 'next-auth/react';
 import { useForm } from 'react-hook-form';
-import NextLink from 'next/link';
 import { useRouter } from 'next/router';
 import Link from 'next/link';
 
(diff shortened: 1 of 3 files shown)


packages/create-bison-app/template/src/components/auth/LoginForm.tsx
```diff
@@ -91,11 +90,11 @@ export function LoginForm() {
         </PasswordField>
       </div>
       <div className="flex justify-between">
-        <NextLink href="/reset-password">
-          <Button variant="link" size="sm">
```
dennis-campos, line 95: I don't think `NextLink` is necessary anymore. I'm also not seeing it in the "pages" docs

packages/create-bison-app/template/src/components/auth/LoginForm.tsx
```diff
@@ -91,11 +90,11 @@ export function LoginForm() {
...
+        <Link href="/reset-password">
+          <Button variant="link" size="sm" type="button">
```
dennis-campos, line 94: If the user hits enter on their keyboard, it redirected them to `/reset-password` instead of signing in. The workaround is having `type=button` attribute. This is necessary when there are multiple buttons in a form.

packages/create-bison-app/template/src/layouts/LoggedIn.tsx
```diff
@@ -12,11 +11,8 @@ interface Props {
 }
 
 export function LoggedInLayout({ children }: Props) {
-  const router = useRouter();
-
   async function handleLogout() {
-    signOut();
-    await router.replace('/login');
```
dennis-campos, line 19: This caused a flash to `/login` page and redirected the user back to `/`

kgajera (approved):

cullylarson (approved):

